PR #13, opened 2025-01-23
Build pg_documentdb on PostgreSQL 17

marcoslot:
A few small changes to get it building on PG17.


visridha:
I would hold on to this for a moment - would prefer to also enable the build actions to build PG17 along with this change

visridha (approved):

patty-chow (approved):
Looks great!

PR #14, opened 2025-01-24
Fix compiler warnings

marcoslot:
was getting some compiler warnings from clang (used when compiling PG with --with-llvm):

```
src/aggregation/bson_project.c:1151:48: warning: use of GNU old-style field designator extension [-Wgnu-designator]                                                                           
        ParseAggregationExpressionContext context = { allowRedactVariables: true };                                                                                                           
                                                      ^~~~~~~~~~~~~~~~~~~~~                

rc/aggregation/bson_densify.c:795:4: warning: expression which evaluates to zero treated as a null pointer constant of type 'pgbson *' [-Wnon-literal-null-conversion]                       
                        PG_RETURN_NULL();                                                                                                                                                     
                        ^~~~~~~~~~~~~~~~      
```

### DIFF
--- a/pg_documentdb/src/aggregation/bson_densify.c
+++ b/pg_documentdb/src/aggregation/bson_densify.c
@@ -294,6 +294,11 @@ Datum
 bson_densify_range(PG_FUNCTION_ARGS)
 {
 	pgbson *result = DensifyPartitionCore(fcinfo, DENSIFY_TYPE_RANGE);
+	if (result == NULL)
+	{
+		PG_RETURN_NULL();
+	}
+
 	PG_RETURN_POINTER(result);
 }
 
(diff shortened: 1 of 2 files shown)


safern (approved):
Thank you, @marcoslot

jayanta-mondal (approved):

sivethe:
@marcoslot: it looks like these changes are failing our tests.  Please take a look

pg_documentdb/src/aggregation/bson_densify.c
```diff
@@ -792,7 +792,7 @@ DensifyPartitionCore(PG_FUNCTION_ARGS, DensifyType type)
 			 *     - Empty collection.
 			 *     - Filtered collection with 0 rows.
 			 */
-			PG_RETURN_NULL();
+			return NULL;
```
visridha, line 805: So previously this would set `fcinfo->isnull = true;` since this takes the fcinfo.
Now callers get `PG_RETURN_POINTER(Datum(0)) ` but we no longer set `fcinfo->isnull`
Is that gonna cause issues?

Should we just change the return type of this method to return a Datum and have the callers do PG_RETURN_DATUM()

marcoslot: Probably nicer to check for NULL in bson_densify_range and do PG_RETURN_NULL there.

diipak-bisht (approved):

visridha (approved):

Issue #40, opened 2025-01-29
`scripts/start_oss_server.sh -c` breaks second time onwards


diipak-bisht:
**Describe the bug**
The documentation suggests to use  `scripts/start_oss_server.sh` script to start the postgresql server with documentdb extension loaded, but -c option is broken 2nd time onwards on a fresh setup

**Environment**
- Operating system and version - Ubuntu 22.04
- PostgreSQL version - 16
- Architecture

**Reproduction Steps**
On a fresh github clone of the repo, after building and making the project run the below script twice.
`scripts/start_oss_server.sh -c`

**Expected behavior**
PostgreSQL server should start without any errors

**Actual behavior**
ERROR:  function public.geometry_gist_union_2d(bytea, internal) does not exist

**Additional context**

visridha:
This should be resolved in the next sync.

Issue #32, opened 2025-01-28
update assignment


maddin200:
/pg_documentdb_core/src/io/pgbson_utils.c line 789
Code:
static void
AddInt64ToValue(bson_value_t *current, int64_t value, bool *overflowedFromInt64)
{
	if (current->value_type == BSON_TYPE_DOUBLE)
	{
		overflowedFromInt64 = false; // <-- should be *overflowedFromInt64 =

diipak-bisht:
nice catch!! Seems like this can be removed for `DOUBLE` as this is not actually an overflow from int

diipak-bisht:
Or should be assigned at least `false` for double and decimal128 both so that callers can't set assumptions

PR #42, opened 2025-01-30
DocumentDB sync 20250130

sivethe:


sivethe:
Resolves #40 
Resolves #32

PR #53, opened 2025-02-04
Parse cpuinfo on aarch64 machines

jamesbroadhead:
- Note this will reduce the number of threads used on x86 to 1:processor rather than 1:core


visridha (approved):

safern (approved):

safern:
Thanks!

Issue #57, opened 2025-02-12
[Performance] Aggregation pipeline merge stage and hash compare optimizations


fonsecamar:
Ref.: File bson_aggregation_output_pipeline.c -> method bson_dollar_merge_handle_when_matched

Currently, $merge using default whenMatched=merge option iterates target and source document attributes in 2 different loops.

Performance suggestion: perform target and source merge in a single loop.
Merging 2 documents with 20 elements each would result in 20 iterations instead of 40. New Max Iterations = MAX(# of elements at source, # of elements at target).

Ref.: File hashset_utils.c -> methods PgbsonElementHashEntryCompareFunc and PgbsonElementOrderedHashCompareFunc

Currently, the hash compare functions are calculating min length between elements and calling strncmp for every call.

Performance suggestion: first compare lengths and return if not zero, and if not, then call strncmp to compare element paths.

lichoil:
issue got fixed in https://github.com/microsoft/documentdb/pull/58.

PR #58, opened 2025-02-12
Aggregation pipeline merge stage and hash compare optimizations

fonsecamar:
Ref #57

- Updating merge algorithm to use a single loop to iterate source and target documents.
- Updating hash compare functions to call memcmp only in case path lengths are equal.

Both regression tests passed 100%.

### DIFF
--- a/pg_documentdb_core/src/utils/hashset_utils.c
+++ b/pg_documentdb_core/src/utils/hashset_utils.c
@@ -80,14 +80,14 @@ PgbsonElementHashEntryCompareFunc(const void *obj1, const void *obj2, Size objsi
 	const PgbsonElementHashEntry *hashEntry1 = obj1;
 	const PgbsonElementHashEntry *hashEntry2 = obj2;
 
-	int minLength = Min(hashEntry1->element.pathLength, hashEntry2->element.pathLength);
-	int result = strncmp(hashEntry1->element.path, hashEntry2->element.path, minLength);
-
-	if (result == 0)
-	{
-		return hashEntry1->element.pathLength - hashEntry2->element.pathLength;
+	// First, compare the lengths
+	int lenDiff = hashEntry1->element.pathLength - hashEntry2->element.pathLength;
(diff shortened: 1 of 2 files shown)


fonsecamar:
@sivethe

pg_documentdb_core/src/utils/hashset_utils.c
```diff
@@ -80,14 +80,14 @@ PgbsonElementHashEntryCompareFunc(const void *obj1, const void *obj2, Size objsi
...
 	const PgbsonElementHashEntry *hashEntry2 = obj2;
 
-	int minLength = Min(hashEntry1->element.pathLength, hashEntry2->element.pathLength);
-	int result = strncmp(hashEntry1->element.path, hashEntry2->element.path, minLength);
-
-	if (result == 0)
-	{
-		return hashEntry1->element.pathLength - hashEntry2->element.pathLength;
+	// First, compare the lengths
+	int lenDiff = hashEntry1->element.pathLength - hashEntry2->element.pathLength;
```
diipak-bisht, line 85: length comparison could be incorrect here for comparing the path strings to be greater or not right?

e.g. 
entry1.path => 'abcdef'
entry2.path => `az`

`entry1` surely has longer string pathlength but it isn't necessarily greater than `entry2`

fonsecamar: @diipak-bisht , as we discussed, HashCompareFunc semantics according to PGSQL docs is to return 0 for match and non-0 for no match: https://doxygen.postgresql.org/hsearch_8h_source.html#l00029.

I've updated the compare functions comments to reflect that.

safern:
Hello @fonsecamar - just wondering if you could share the perf results that @LordParag asked for so that we can move this pull request forward.

safern:
@fonsecamar - this has been stale for some time and it also has conflicts now. I'm gonna close the pull request, feel free to reopen/open a new one when you have some time to look at this. 

Thank you in advance!

PR #74, opened 2025-02-21
Fix PostgreSQL 17 Building Failure by Hiding query->mergeUseOuterJoin 

Vonng:
Related Issue: https://github.com/microsoft/documentdb/issues/73

`query->mergeUseOuterJoin` is removed in postgres 17

- [PG16: https://github.com/postgres/postgres/blob/REL_16_8/src/include/nodes/parsenodes.h#L186](https://github.com/postgres/postgres/blob/REL_16_8/src/include/nodes/parsenodes.h#L186)
- [PG17: https://github.com/postgres/postgres/blob/REL_17_4/src/include/nodes/parsenodes.h#L186](https://github.com/postgres/postgres/blob/REL_17_4/src/include/nodes/parsenodes.h#L186)

The following marco fix the build issue against PostgreSQL 17

```c
	#if PG_VERSION_NUM < 170000
	query->mergeUseOuterJoin = true;
	#endif
```

You may have to replace it with the PG 17 new fields in the future.

```c
	int  mergeTargetRelation pg_node_attr(query_jumble_ignore);

	/* join condition between source and target for MERGE */
	Node *mergeJoinCondition;
```

But for now, at least it can pass the building process for PG 17.

### DIFF
--- a/pg_documentdb/src/aggregation/bson_aggregation_output_pipeline.c
+++ b/pg_documentdb/src/aggregation/bson_aggregation_output_pipeline.c
@@ -1272,7 +1272,9 @@ AddTargetCollectionRTEDollarMerge(Query *query, MongoCollection *targetCollectio
 	RangeTblEntry *existingrte = list_nth(query->rtable, 0);
 	query->rtable = list_make2(rte, existingrte);
 	query->resultRelation = 1;
+	#if PG_VERSION_NUM < 170000
 	query->mergeUseOuterJoin = true;
+	#endif
 	query->targetList = NIL;
 }
 


safern:
@LordParag could you work with @Vonng to update accordingly for merge to work with the new Postgres 17 fields?

safern:
@Vonng thank you for working on this. Unfortunately this was already taken care as this was blocking some of our product testing and such, so I'll close this PR, but I appreciate you contributing, I hope we see another contribution from you on the future. 

Also thank you for reporting the issue 😄 

Should be fixed with: https://github.com/microsoft/documentdb/pull/78

PR #84, opened 2025-02-26
Fix timezone string corruption caused by stack lifetime.

buriedpot:
**Describe the bug**

bson_expression_date_operators.c:6941

Image

bson_expression_date_operators.c:7064

Image

In these two places, the code allocates string space on the stack and assigns it to dateFromParts->timezone later: 
Image

**Environment**
- Linux
- PostgreSQL 16
- Architecture

**Reproduction Steps**

compilation: gcc -fstack-protector-strong -O0
run test case: bson_aggregation_date_operators_tests

```PostgreSQL
postgres=# select * from bson_dollar_project('{}','{"result": {"$dateFromString": {"dateString": "2021-01-11T12:53:12.506+130", "format": "%Y-%m-%dT%H:%M:%S.%L%Z"} } }');
ERROR:  unrecognized time zone identifier: "t���"
```

**Solution**

Maybe using palloc0 is reasonable instead of allocating string on stack.

### DIFF
--- a/pg_documentdb/src/operators/bson_expression_date_operators.c
+++ b/pg_documentdb/src/operators/bson_expression_date_operators.c
@@ -6938,7 +6938,7 @@ ValidateOffsetMinutes(char *dateString, int sizeOfDateString, int *indexOfDateSt
 	int hours = totalMins / 60;
 	int minutes = totalMins % 60;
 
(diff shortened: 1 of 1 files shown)


safern (approved):
I think this makes sense, thank you @buriedpot

PR #91, opened 2025-03-03
Fix use validator on stack after returning.

buriedpot:
**Describe the bug**

pg_documentdb/src/metadata/collection.c:1905, the function ParseAndGetValidatorSpec returns a variable 'validator' allocated on stack.
```c
//...
	EnsureTopLevelFieldType(validatorName, iter, BSON_TYPE_DOCUMENT);
	const bson_value_t *validator = bson_iter_value(iter);

	/* Large and overly complex validation rules can impact database performance, especially during write operations. */
	if (validator->value.v_doc.data_len > (uint32_t) MaxSchemaValidatorSize)
//...
```

pg_documentdb/src/commands/create_collection_view.c:382, the function ParseCreateSpec assign the pointer to spec->validator, which will be used after the function returning.

```c
else if (strcmp(key, "validator") == 0)
{
	spec->validator = ParseAndGetValidatorSpec(&createIter, "create.validator",
										   hasSchemaValidationSpec);
}
```

**Environment**
- Linux
- PostgreSQL 16
- Architecture

**Reproduction Steps**

compilation: gcc -fstack-protector-strong -O0
run test case: schema_validation_insert, schema_validation

```PostgreSQL
postgres=# SELECT documentdb_api.create_collection_view('schema_validation', '{ "create": "col", "validator": {"$jsonSchema": {"bsonType": "object", "properties": {"a": {"bsonType": "int"}}}}, "validationLevel": "strict", "validationAction": "error"}');
NOTICE:  creating collection
ERROR:  expected a document to create a bson object
```

**Solution**

Use palloc0.

### DIFF
--- a/pg_documentdb/src/commands/create_collection_view.c
+++ b/pg_documentdb/src/commands/create_collection_view.c
@@ -378,8 +378,10 @@ ParseCreateSpec(Datum databaseDatum, pgbson *createSpec, bool *hasSchemaValidati
 		}
 		else if (strcmp(key, "validator") == 0)
 		{
-			spec->validator = ParseAndGetValidatorSpec(&createIter, "create.validator",
-													   hasSchemaValidationSpec);
+			bson_value_t *validator = palloc0(sizeof(bson_value_t));
+			*validator = *(ParseAndGetValidatorSpec(&createIter, "create.validator",
+													hasSchemaValidationSpec));
+			spec->validator = validator;
 		}
 		else if (strcmp(key, "validationLevel") == 0)
 		{


pg_documentdb/src/commands/create_collection_view.c
```diff
@@ -378,8 +378,10 @@ ParseCreateSpec(Datum databaseDatum, pgbson *createSpec, bool *hasSchemaValidati
 		}
 		else if (strcmp(key, "validator") == 0)
 		{
-			spec->validator = ParseAndGetValidatorSpec(&createIter, "create.validator",
-													   hasSchemaValidationSpec);
+			bson_value_t *validator = palloc0(sizeof(bson_value_t));
```
safern, line 381: Could we avoid the palloc0 changing `ParseAndGetValidatorSpec` to take a pointer and assign the value to such pointer?

Then we could call the function like:
`ParseAndGetValidatorSpec(&createIter, "create.validator", hasSchemaValidationSpec, &spec->validator)`

buriedpot: Thank you for your review 😊

Do you mean that we should still allocate validator on stack or we should allocate validator on heap in function ParseAndGetValidatorSpec instead of ParseCreateSpec?

As you can see at `create_collection_view.c:154`, after function ParseCreateSpec returning, spec->validator will be still used, so I think that using palloc is necessary. I can put the palloc into function ParseAndGetValidatorSpec.

buriedpot: Or you want to use palloc instead of palloc0?

I think using palloc is indeed more reasonable.

safern: What I mean is that spec in create_collection_view is palloc'd already and not in the stack. So we could just change the signature of the `ParseAndGetValidatorSpec` function to accept a `bson_value_t *` instead of returning `bson_value_t`.

Then it will populate the `bson_value_t *` passed in with the parsed validator spec. 

Then we call is like this:
```
ParseAndGetValidatorSpec(&createIter, "create.validator", hasSchemaValidationSpec, &spec->validator)
```

buriedpot: Thank you for reviewing 😊

We can start an psql connection and `gdb --tui attach <backend_pid>` and create a breakpoint at `src/metadata/collection.c:1905`. Then you can execute query `SELECT documentdb_api.create_collection_view('schema_validation', '{ "create": "col", "validator": {"$jsonSchema": {"bsonType": "object", "properties": {"a": {"bsonType": "int"}}}}, "validationLevel": "strict", "validationAction": "error"}');` in this psql connection. After the statement `const bson_value_t *validator = bson_iter_value(iter);` executing, we can execute `p validator` in gdb, and then we can see the following content:

```
(gdb) p validator
$5 = (const bson_value_t *)
(gdb)
0x7ffdc2491db0
```
We can execute `info proc mapping` in gdb, and it will tell us that stack and heap memory range:

```
Ox7ffdc2292000 Ox7ffdc2496000 0x204000 0x0 [stack]
```

If you compile documentdb with `gcc -O0 -fstack-protector-strong`, we can execute `c` in gdb and then an error will be logged out. So the validator points to an memory area on stack.

We can use test cases schema_validation_insert and schema_validation to reproduct the error.

buriedpot: Can you provide the libbson version recommended? I can try to compile documentdb again with the libbson version.

buriedpot: Thank you for your advice, I will try to make some modifications.

buriedpot: Thank you for your review.
The validator field in spec is a pointer, so palloc is inevitable.

I can put palloc in function ParseAndGetValidatorSpec instead of ParseCreateSpec.

The core problem is that the function `bson_iter_value` will return a pointer which points to a memory area on stack.

visridha: `bson_value_t` is a struct right? - so if we have spec's validator be a `bson_value_t` (instead of a `const bson_value_t *`) then when calling `ParseAndGetValidatorSpec` as called above, we can have something like

```
specValidator = *bson_iter_value(&iterator)
```

buriedpot: Ok, I will make some modifications. Thank you for reviewing.

safern: Hi, @buriedpot -- where you able to do the adjustment suggested here?

safern:
@buriedpot friendly ping, will you be able to finish this? Thank you in advance.

jiahu2:
Hi @buriedpot have you encountered this issue recently? I haven’t been able to reproduce it, and I’d love to work with you to resolve it.

jiahu2:
Hi, the issue has been resolved in June, will close this PR.

PR #92, opened 2025-03-04
[Test] fix: use normalized results file in test cases

buriedpot:
In distributed scenarios, we should compare the expected file with result files normalized by normalize.sed.


buriedpot:
@microsoft-github-policy-service agree

PR #93, opened 2025-03-04
[Test] fix: use normalized expected files in test cases

buriedpot:
In distributed scenarios, we should compare the expected file with result files normalized by normalize.sed, but the expected file has some unnormalized content. We fixed this by replace strings like "123_1" to "X_X".

### DIFF
--- a/internal/pg_documentdb_distributed/src/test/regress/expected/bson_aggregation_pipeline_stage_densify.out
+++ b/internal/pg_documentdb_distributed/src/test/regress/expected/bson_aggregation_pipeline_stage_densify.out
@@ -687,7 +687,7 @@ EXPLAIN (COSTS OFF, VERBOSE ON) SELECT document FROM bson_aggregation_pipeline('
 ---------------------------------------------------------------------
  Custom Scan (Citus Adaptive)
    Output: remote_scan.document
-   ->  Distributed Subplan 54_1
+   ->  Distributed Subplan X_X
          ->  Custom Scan (Citus Adaptive)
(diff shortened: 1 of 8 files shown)


safern:
Hello @buriedpot -- thanks for your contribution.

The reason why we don't check in the normalized output is because the way our test infrastructure works is by running normalize.sed first on the files and writint that into a `.modified` file.

And then we do the diff between the 2 `.modified` files (the one in expected and in the result). 

https://github.com/microsoft/documentdb/blob/f237956efbf871176b9cd6b6b85f694c2f7fed4d/internal/pg_documentdb_distributed/src/test/regress/bin/diff#L49C2-L66C110

That's why the tests don't fail.

Am I missing something that you saw on to why we need to update the checked in out files as normalized files?

buriedpot:
> Hello @buriedpot -- thanks for your contribution.
> 
> The reason why we don't check in the normalized output is because the way our test infrastructure works is by running normalize.sed first on the files and writint that into a `.modified` file.
> 
> And then we do the diff between the 2 `.modified` files (the one in expected and in the result).
> 
> https://github.com/microsoft/documentdb/blob/f237956efbf871176b9cd6b6b85f694c2f7fed4d/internal/pg_documentdb_distributed/src/test/regress/bin/diff#L49-L66
> 
> That's why the tests don't fail.
> 
> Am I missing something that you saw on to why we need to update the checked in out files as normalized files?

Thank you for reviewing. I agree with you and I will close this PR.

PR #98, opened 2025-03-07
Replace deprecated `bson_as_json` with `bson_as_legacy_extended_json`

wangxuqi:
As `bson_as_json`  is deprecated. Replace it with `bson_as_legacy_extended_json`.


diipak-bisht:
I think we should also upgrade libbson if we want to remove the deprecated functions, documentdb still uses 1.28

wangxuqi:
> I think we should also upgrade libbson if we want to remove the deprecated functions, documentdb still uses 1.28

OK, get. Close it first.  Reopen if  libbson upgrade to higher version.

Issue #97, opened 2025-03-07
Source build requires internal intelmath header


mbanck:
**Describe the bug**

Building documentdb on Debian unstable with the system-provided `libintelrdfpmath-dev`  package results in a build failure:

```
/usr/bin/clang-19 -Wno-ignored-attributes -fno-strict-aliasing -fwrapv -fexcess-precision=standard -Wno-unused-command-line-argument -Wno-compound-token-split-by-macro -Wno-format-truncation -O2  -DBSON_STATIC -I/usr/include/libbson-1.0  -I/usr/include/postgresql   -Isrc -Iinclude -Ibuild/include/ -I. -I./ -I/usr/include/postgresql/17/server -I/usr/include/postgresql/internal  -Wdate-time -D_FORTIFY_SOURCE=2 -D_GNU_SOURCE -I/usr/include/libxml2  -flto=thin -emit-llvm -c -o src/types/decimal128.bc src/types/decimal128.c
src/types/decimal128.c:21:10: fatal error: 'bid_internal.h' file not found
   21 | #include <bid_internal.h>
      |          ^~~~~~~~~~~~~~~~
1 error generated.
```

This is because `decimal128.c` includes an internal header that is (rightfully) not shipped in the Debian package at https://github.com/microsoft/documentdb/blob/main/pg_documentdb_core/src/types/decimal128.c#L21

If I remove that include, I get the following build error:

```
src/types/decimal128.c: In function ‘IsDecimal128Infinity’:
src/types/decimal128.c:432:61: error: ‘SINFINITY_MASK64’ undeclared (first use in this function)
  432 |                 *isPositiveInfinity = (HIGH_BITS(value)) != SINFINITY_MASK64;
      |                                                             ^~~~~~~~~~~~~~~~
src/types/decimal128.c:432:61: note: each undeclared identifier is reported only once for each function it appears in
src/types/decimal128.c: In function ‘SetDecimal128NaN’:
src/types/decimal128.c:466:42: error: ‘NAN_MASK64’ undeclared (first use in this function)
  466 |         value->value.v_decimal128.high = NAN_MASK64;
      |                                          ^~~~~~~~~~
src/types/decimal128.c: In function ‘SetDecimal128NegativeInfinity’:
src/types/decimal128.c:477:42: error: ‘SINFINITY_MASK64’ undeclared (first use in this function)
  477 |         value->value.v_decimal128.high = SINFINITY_MASK64;
[...]

PR #99, opened 2025-03-07
Avoid including internal libintelmath header.

mbanck-ntap:
This removed the include of <bid_internal.h> and adds the three MASK64 definitions that are used in the code.

Fixes: #97


lichoil:
hi @mbanck-ntap ,may I know the reason why removing this header file out and put MASK64 definitions in stead?

mbanck:
> @lichoil I think it is to be able to compile in debian.

Exactly.

diipak-bisht (approved):

diipak-bisht:
I think this is fine change if internal headers are not available in debian unstable @safern. Thanks for contributing this @mbanck

visridha (approved):

PR #105, opened 2025-03-14
Add dockerfile for pre-build image

fatmelon:
Add files for pre-build image and start services script.

- Navigate to the repo and build this image with:
    ```bash
    docker build . -f .\.github\containers\Build-Ubuntu\Dockerfile_prebuild -t <image-tag> 
    ```
- The image is now available for:

    ```bash
    docker pull mcr.microsoft.com/cosmosdb/ubuntu/documentdb-oss:20.04
    ```

- Base image is built from `.github\containers\Build-Ubuntu\Dockerfile`

### DIFF
--- a/.github/containers/Build-Ubuntu/Dockerfile_prebuild
+++ b/.github/containers/Build-Ubuntu/Dockerfile_prebuild
@@ -1,0 +1,11 @@
+FROM mcr.microsoft.com/cosmosdb/ubuntu/documentdb-oss:base
+
+WORKDIR /home/documentdb/code  
+
+COPY . /home/documentdb/code  
+
+RUN git config --global --add safe.directory /home/DocumentDB/code
+RUN make  && sudo make install
+  
+# Set the entry point to run the new script at container startup  
+ENTRYPOINT ["/home/documentdb/code/scripts/start_services.sh"]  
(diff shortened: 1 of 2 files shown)


.github/containers/Build-Ubuntu/Dockerfile_prebuild
```diff
@@ -0,0 +1,11 @@
...
+
+WORKDIR /home/documentdb/code  
+
+COPY . /home/documentdb/code  
+
+RUN git config --global --add safe.directory /home/DocumentDB/code
+RUN make  && sudo make install
+  
+# Set the entry point to run the new script at container startup  
+ENTRYPOINT ["/home/documentdb/code/scripts/start_services.sh"]
```
lichoil: CMD ["bash", "-c", "./scripts/start_oss_server.sh & tail -f /dev/null"]
it's better keep container running in background and get back to previous terminal session, and needn't start psql. and avoid adding another script.

scripts/start_services.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
lichoil: don't need this file. can put the server start command in dockerfile.

.github/containers/Build-Ubuntu/Dockerfile_prebuild
fatmelon: updated

scripts/start_services.sh
fatmelon: updated

lichoil (approved):
pls update the docker image in mcr using this dockerfile build after pr merged.

PR #111, opened 2025-03-25
Fix handling of explicit `maxTimeMS` zero values

AlekSi:
Explicit zero `maxTimeMS` value should be treated as unset.

Tested by FerretDB's integration tests.


AlekSi:
@visridha PTAL

visridha (approved):

Issue #94, opened 2025-03-05
[Feature Request] Provie ready-to-use files or packages (e.g. .rpm or .deb) for DocumentDB


xenuser:
**Purpose of the feature.**
Make it easy to install DocumentDB without the need to compile it on your own

**Describe the solution you'd like**
Sometimes the Docker image of DocumentDB can not be used, and self-building is not an option due to complex enterprise proxy limitations. Having a ready-to-use "file" or package (e.g. an RPM) that can be grabbed from the project's releases page would help to easily install DocumentDB on top of Postgres.

We also have this with FerretDB, so if possible, we should also have it here with DocumentDB.

This would also help if the available container image can not be used and needs to be re-built, e.g. because the base image needs to be SUSE or Red Hat.

**Describe alternatives you've considered**
A ready-to-use image based on Red Hat's UBI would also help some enterprises to just use DocumentDB.

**Additional context**
None

xenuser:
> FWIW, we (FerretDB) provide `.deb` packages for Debian and Ubuntu on our release pages, like [that one](https://github.com/FerretDB/documentdb/releases/tag/v0.102.0-ferretdb-2.0.0). Sources are [there](https://github.com/FerretDB/documentdb/tree/ferretdb/ferretdb_packaging).

This is an excellent tip. Thanks! I checked out the released files, but I was not aware of DocumentDB being a part of the DEB package.

safern:
Related: #54

mbanck-ntap:
> FWIW, we (FerretDB) provide `.deb` packages for Debian and Ubuntu on our release pages, like [that one](https://github.com/FerretDB/documentdb/releases/tag/v0.102.0-ferretdb-2.0.0). Sources are [there](https://github.com/FerretDB/documentdb/tree/ferretdb/ferretdb_packaging).

Why do you need this? https://github.com/FerretDB/documentdb/blob/ferretdb/ferretdb_packaging/debian_files/rules#L8-L10

AlekSi:
> Why do you need this? https://github.com/FerretDB/documentdb/blob/ferretdb/ferretdb_packaging/debian_files/rules#L8-L10

It is a part of the package testing. Just installing a package does not create this user.

Issue #23, opened 2025-01-26
support ARM architecture


GithubArray:
**Purpose of the feature.**
In the StackOverflow Community, one of the most-requested features for Azure CosmosDB emulator is the support for Apple Silicon chip: https://stackoverflow.com/questions/75191130/cosmos-db-emulator-installation-on-mac/75191141#75191141

I would like to confirm whether this project is officially supported for Apple Silicon chip. If not, will it be supported in future releases?

**Describe the solution you'd like**
I have tried to build the docker image on my Apple M1 machine but had an error. Here is the excerpt of the exception message:
```
105.5 
105.5 This program built for x86_64-pc-linux-gnu 
105.5 Report bugs to <bug-make@gnu.org>
------
Dockerfile:63
--------------------
  61 |     # Install postgres
  62 |     COPY scripts/install_setup_postgres.sh /tmp/install_setup/
  63 | >>> RUN [ "bin/bash", "-c", "/tmp/install_setup/install_setup_postgres.sh -d \"/usr/lib/p
ostgresql/${PG_VERSION}\" $POSTGRES_INSTALL_ARG -v ${PG_VERSION}" ]                             
  64 |     
  65 |     # Install RUM from source
--------------------
ERROR: failed to solve: process "bin/bash -c /tmp/install_setup/install_setup_postgres.sh -d \"/
usr/lib/postgresql/${PG_VERSION}\" $POSTGRES_INSTALL_ARG -v ${PG_VERSION}" did not complete succ
essfully: exit code: 2  
```

I noticed that there is a line: `This program built for x86_64-pc-linux-gnu` 

**Describe alternatives you've considered**
I have tried to check out the [Roadmap list](https://github.com/orgs/microsoft/projects/1407/views/1) in the README, but the link is 404 there.

AlekSi:
arm64 support was requested by the FerretDB community almost immediately.

AlekSi:
> I have tried to check out the [Roadmap list](https://github.com/orgs/microsoft/projects/1407/views/1) in the README, but the link is 404 there.

That part is fixed now.

ringods:
Please make sure this also runs on Raspberry Pi 4 (Architecture `linux/arm/v8`, more specifically v8.0).

AlekSi:
> @AlekSi regarding the Roadmap, I see multiple issues listed as You can't see this item
> Can you have a look at that?

[...]

PR #125, opened 2025-03-27
Add scripts to generate deb packages and set up new workflow

shuaitian-git:
fix [#94](https://github.com/microsoft/documentdb/issues/94)


.github/workflows/build_packages.yml
```diff
@@ -0,0 +1,62 @@
...
+  group: build-packages-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - 'main'
+    paths-ignore:
+      - 'docs/**'
```
fatmelon: we don't have docs folder currently

packaging/build_packages.sh
```diff
@@ -0,0 +1,157 @@
...
+cd $repo_root
+
+echo "Building packages for OS: $OS, PostgreSQL version: $PG, DOCUMENTDB version: $DOCUMENTDB_VERSION"
+echo "Output directory: $abs_output_dir"
+
+# Create the output directory if it doesn't exist
+mkdir -p $abs_output_dir
+
+# Build the Docker image while showing the output to the console
+docker build --platform linux/amd64 -t documentdb-build-packages:latest -f packaging/Dockerfile_build_deb_packages \
```
fatmelon: nit: maybe use the combination of pg_version and os_verison as the image tag？

lichoil: i think we need provide both amd and arm version packages?

safern:
@shuaitian-git could you create a sub issue from the original issue in order to publish the deb packages?

packaging/debian_files/changelog
```diff
@@ -0,0 +1,5 @@
+documentdb (DOCUMENTDB_VERSION) Unreleased; urgency=low
```
safern, line 1: should we just generate this file on the fly from the changelog.md file?

packaging/debian_files/control
```diff
@@ -0,0 +1,10 @@
+Source: documentdb
+Section: database
+Priority: optional
+Build-Depends: debhelper (>= 11), postgresql-server-dev-POSTGRES_VERSION
+Maintainer: Shuai Tian <shuaitian@microsoft.com>
```
safern, line 5: would it be better to have an org for this? What are other products in MSFT doing?

.github/workflows/regress_tests.yml
```diff
@@ -133,9 +134,9 @@ jobs:
           export LC_CTYPE=en_US.UTF-8
           make check
 
-      - name: Citus Indent
-        run: |
-          citus_indent --check || (echo ""; echo "citus_indent failed, refer to CONTRIBUTING.md on how to use citus_indent"; exit 1)
+      # - name: Citus Indent
```
safern: why?

packaging/build_packages.sh
```diff
@@ -0,0 +1,157 @@
...
+cd $repo_root
+
+echo "Building packages for OS: $OS, PostgreSQL version: $PG, DOCUMENTDB version: $DOCUMENTDB_VERSION"
+echo "Output directory: $abs_output_dir"
+
+# Create the output directory if it doesn't exist
+mkdir -p $abs_output_dir
+
+# Build the Docker image while showing the output to the console
+docker build --platform linux/amd64 -t documentdb-build-packages-$OS-pg$PG:latest -f packaging/Dockerfile_build_deb_packages \
```
AlekSi: It would be great to use [`docker buildx build`](https://docs.docker.com/reference/cli/docker/buildx/build/) with `--platform linux/amd64,linux/arm64` to build arm64 packages, too (#23)

.github/workflows/regress_tests.yml
shuaitian-git: was testing it

packaging/debian_files/changelog
shuaitian-git: debian/changelog has its own structure and format, if we write a script for conversion, we need to be strict on the format of CHANGELOG.md every time we update it. I think manual copy may be better.

packaging/debian_files/control
shuaitian-git: I searched and didn't find any control file having org section. We can add more folks with an uploaders section.

Typical usage:
https://github.com/microsoft/lsvmtools/blob/2d7e586e408947e7c905145550f5ae4cc1f2bc52/deb/debian/control#L5

shuaitian-git:
> @shuaitian-git could you create a sub issue from the original issue in order to publish the deb packages?

created #139

shuaitian-git:
changed to use ubuntu-24.04-arm runner for arm64 builds

lichoil (approved):

packaging/build_packages.sh
shuaitian-git: good catch, need to remove the --platform arg

shuaitian-git: This arg is already removed from latest commit

lichoil (approved):

Issue #96, opened 2025-03-06
Tarball from tags do not build


mbanck:
**Describe the bug**

Trying to build the v0.101-0 tag tarballresults in (besides potentially other issues) in an unhelpful error:

```
fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
make[1]: *** No rule to make target '/index', needed by 'build/include/metadata/documentdb_extension_version.h'.  Stop.
make[1]: Leaving directory '/build/documentdb-meFFdU/documentdb-0.101-0/pg_documentdb'
make: *** [Makefile:11: .DEFAULT] Error 2
```

It took me a while to figure out what is going on here, but this `/index` is coming from a non-assigned variable from `Makefile.versions`:

```
GIT_INDEX_DIR=$(shell git rev-parse --git-dir)/index
```

If I take that out, it fails further down the road in `scripts/generate_extension_version.sh`

I would say the release tarballs should just ship a pre-generated `documentdb_extension_version.h`

**Environment**
- Operating system and version Debian unstable
- PostgreSQL version 17
- Architecture amd64

**Reproduction Steps**

Run `make`  from the release tarball.

**Expected behavior**

Build succeeds

**Actual behavior**

**Additional context**

safern:
@mbanck - thanks for reporting the issue.

We definitely need to improve building the tarball from the releases providing a configure -> make solution, which we have planned to do, so thanks for opening this issue.

In the meantime an alternative solution is to clone the repository and checkout the release tag and build following the README steps, would that work to unblock you for now? 

@lichoil might be good to plan and assign this accordingly

AlekSi:
One another option is to actually upload tarballs with all generated files. "Source code (tar.gz)" is generated on the fly by GitHub and may change over time, breaking checksums and such: https://docs.github.com/en/repositories/working-with-files/using-files/downloading-source-code-archives#stability-of-source-code-archives.

PR #132, opened 2025-04-02
[Fix] Update Makefile.versions to avoid build failure with tarball from tags 

fatmelon:
In this pull request, update the `Makefile.versions` and `scripts/generate_extension_version.sh` to avoid build failures when using a tarball from tags.

In release tarball there is no .git/index file, so we need to check if the file exists before using it. If the file does not exist, we will set some variables to unknown and use the version from the control file.


safern:
@fatmelon could you please add the issue to the description so that it is closed automatically?

fixes: #96

PR #136, opened 2025-04-03
enforce cmake version in sql tests workflow 

### LINKED ISSUE #137: cmake non compatibility error
**Describe the bug**
The regress test pipeline is failing due to a cmake version error:

CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
-- Configuring incomplete, errors occurred!

shuaitian-git:
This is a workaround to quickly fix the cmake version non-compatible error in the regress test workflow.

The citus indent tool is depending on uncrustify, for which we are depending on an old version with cmake_minimum_required set to 2.8.

resolves #137

### DIFF
--- a/.github/workflows/regress_tests.yml
+++ b/.github/workflows/regress_tests.yml
@@ -41,7 +41,7 @@ jobs:
         run: |
           cmake --version
           sudo apt-get update
-          sudo apt-get install -y cmake=3.31.*
+          sudo apt-get install -y cmake=3.31.6
           cmake --version
 
       - name: Install package dependencies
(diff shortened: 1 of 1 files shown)


shuaitian-git:
This is still worth to check-in, to prevent similar issue coming form runner images.

PR #140, opened 2025-04-09
[0.102-1] Cherry picks the changes for the 0.102-1 schema into the release branch

safern:
FerretDB took an early dependency on main when the version was 0.102-0 and now they can't upgrade. We need to make this hotfix version that includes the diff between the commit they took the dependency and the actual 0.102-0 release.

### DIFF
--- a/internal/pg_documentdb_distributed/documentdb_distributed.control
+++ b/internal/pg_documentdb_distributed/documentdb_distributed.control
@@ -1,5 +1,5 @@
 comment = 'Multi-Node API surface for DocumentDB'
-default_version = '0.102-0'
+default_version = '0.102-1'
 module_pathname = '$libdir/pg_documentdb_distributed'
 relocatable = false
 superuser = true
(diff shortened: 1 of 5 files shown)


lichoil (approved):

AlekSi:
I think we discussed in the Discord that we don't actually need that?..  See the thread that ended at https://discord.com/channels/723347736853741589/1355009134700789761/1356465310710108302. We also released FerretDB v2.1.0 already with a clear warning: https://github.com/FerretDB/FerretDB/releases/tag/v2.1.0

safern:
I mean it is not a big deal we already did the work for this and it is just a matter of creating a tag. Let me know if you would like it I can move forward.

I thought what we agreed not to do the change to point to the previous version but releasing this I thought was needed still.

safern:
Looks like ferretdb 2.1 was released already. So will close this and will delete the release branch.

PR #141, opened 2025-04-09
include arm64 in regress test

shuaitian-git:


shuaitian-git:
will change this internally and do a sync

PR #160, opened 2025-04-24
Add prebuild image list

fatmelon:
Add prebuild image list and update README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ docker run -dt mcr.microsoft.com/cosmosdb/ubuntu/documentdb-oss:22.04-PG16-AMD64
 2. If external access is required, run the container with parameter "-e":
 
 ```bash
-docker run -p 9712:9712 -dt mcr.microsoft.com/cosmosdb/ubuntu/documentdb-oss:22.04-PG16-AMD64-0.103.0 -e
+docker run -p 127.0.0.1:9712:9712 -dt mcr.microsoft.com/cosmosdb/ubuntu/documentdb-oss:22.04-PG16-AMD64-0.103.0 -e
 ```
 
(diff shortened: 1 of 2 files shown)


shuaitian-git (approved):

README.md
```diff
@@ -81,40 +81,30 @@ You are all set to work with DocumentDB.
...
 ```bash
-docker run -p 9712:9712 -dt mcr.microsoft.com/cosmosdb/ubuntu/documentdb-oss:20.04
+docker run -dt mcr.microsoft.com/cosmosdb/ubuntu/documentdb-oss:22.04-PG16-AMD64-0.103.0
 ```
 
 2. If external access is required, run the container with parameter "-e":
 
 ```bash
-docker run -p 9712:9712 -dt mcr.microsoft.com/cosmosdb/ubuntu/documentdb-oss:20.04 -e
+docker run -p 9712:9712 -dt mcr.microsoft.com/cosmosdb/ubuntu/documentdb-oss:22.04-PG16-AMD64-0.103.0 -e
```
lichoil: this will open access to postgres from any ip address without passwd which might has risk.

can we use below as sample command ? (only access from host compute to postgres)
docker run -p 127.0.0.1:9712:9712 -dt mcr.microsoft.com/cosmosdb/ubuntu/documentdb-oss:22.04-PG16-AMD64-0.103.0 -e

or if we want keep this and let customer expose it to external any ip, we should add explicit Warning here:
 [!WARNING]
> By doing this, anyone with network access to your machine can connect to PostgreSQL without a password.

fatmelon: updated

lichoil (approved):

PR #195, opened 2025-05-20
Add OpenTelemetry Monitoring to DocumentDB Gateway (Phase 1)

udsmicrosoft:
This PR implements Phase 1 of OpenTelemetry instrumentation for the DocumentDB Gateway. The implementation focuses on two critical metrics that provide immediate operational visibility:

### Cluster Availability Tracking
- Adds metrics to monitor the perceived availability of primary and secondary clusters
- Implemented as gauge metrics with `cluster_id` labels
- Values: 1 (available) or 0 (unavailable)

### Request Routing Monitoring
- Tracks traffic routing between primary and secondary regions
- Implemented as counter metrics with `target_region` and `status` labels
- Provides visibility into traffic shifts during failover events

### The PR includes:
- OpenTelemetry provider implementation
- Telemetry trait integration with the gateway
- Docker-based monitoring stack with:
  - OpenTelemetry Collector
  - Prometheus
  - Grafana dashboards
- Documentation on monitoring setup and usage

This implementation creates a foundation for enhanced observability that will help detect and diagnose failover events and cluster availability issues in multi-region deployments.

### DIFF
--- a/pg_documentdb_gw/src/main.rs
+++ b/pg_documentdb_gw/src/main.rs
@@ -85,7 +85,21 @@ async fn main() {
     .await
     .unwrap();
 
-    run_server(service_context, certificate_options, None, token.clone(), None)
+    let telemetry_provider = Arc::new(documentdb_gateway::open_telemetry_provider::OpenTelemetryProvider::new());
+    
+    let telemetry_clone = telemetry_provider.clone();
+    
+    // Initialize the cluster monitor with the primary pool
+    let cluster_monitor = documentdb_gateway::monitoring::ClusterMonitor::new(
+        system_pool.clone(), 
+        None, // Replace with secondary pool if available
+        telemetry_clone,
+        30
+    );
+    
+    let _monitor_handle = cluster_monitor.start().await;
+    
+    run_server(service_context, certificate_options, Some(telemetry_provider), token.clone(), None)
         .await
         .unwrap();
 }
--- a/pg_documentdb_gw/src/monitoring/mod.rs
+++ b/pg_documentdb_gw/src/monitoring/mod.rs
@@ -1,0 +1,88 @@
+/*-------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation.  All rights reserved.
+ *
+ * src/monitoring/mod.rs
+ *
+ *-------------------------------------------------------------------------
+ */
+
+use std::sync::Arc;
+use std::time::Duration;
+use tokio::time::interval;
+use log::{info, warn, error};
+
+use crate::postgres::Pool;
+use crate::telemetry::TelemetryProvider;
+
+/// ClusterMonitor periodically checks the availability of primary and secondary clusters
+/// and emits metrics via the TelemetryProvider.
+pub struct ClusterMonitor {
+    primary_pool: Arc<Pool>,
+    secondary_pool: Option<Arc<Pool>>,
+    telemetry: Arc<dyn TelemetryProvider>,
+    check_interval: Duration,
+}
+
+impl ClusterMonitor {
+    pub fn new(
+        primary_pool: Arc<Pool>,
+        secondary_pool: Option<Arc<Pool>>,
+        telemetry: Arc<dyn TelemetryProvider>,
+        check_interval_secs: u64,
+    ) -> Self {
+        ClusterMonitor {
+            primary_pool,
+            secondary_pool,
+            telemetry,
+            check_interval: Duration::from_secs(check_interval_secs),
+        }
+    }
+
+    /// Start the cluster monitoring background task
+    pub async fn start(self) -> tokio::task::JoinHandle<()> {
+        tokio::spawn(async move {
+            let mut interval = interval(self.check_interval);
+            
+            loop {
+                interval.tick().await;
+                self.check_cluster_availability().await;
+            }
+        })
+    }
+
+    /// Check the availability of all clusters and emit metrics
+    async fn check_cluster_availability(&self) {
+        // Check primary cluster availability
+        let primary_available = self.check_pool_availability(&self.primary_pool, "primary").await;
+        self.telemetry.update_cluster_availability("primary", primary_available);
+
+        // Check secondary cluster availability if configured
+        if let Some(secondary_pool) = &self.secondary_pool {
+            let secondary_available = self.check_pool_availability(secondary_pool, "secondary").await;
+            self.telemetry.update_cluster_availability("secondary", secondary_available);
+        }
+    }
+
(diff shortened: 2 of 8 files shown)


pg_documentdb_gw/src/main.rs
```diff
@@ -85,7 +85,21 @@ async fn main() {
     .await
     .unwrap();
 
-    run_server(service_context, certificate_options, None, token.clone(), None)
+    let telemetry_provider = Arc::new(documentdb_gateway::open_telemetry_provider::OpenTelemetryProvider::new());
+    
+    let telemetry_clone = telemetry_provider.clone();
+    
+    // Initialize the cluster monitor with the primary pool
```
AndrewKhoma, line 92: nit: it's a system pool for system requests such as get the GUCs from backend

pg_documentdb_gw/src/main.rs
```diff
@@ -85,7 +85,21 @@ async fn main() {
     .await
     .unwrap();
 
-    run_server(service_context, certificate_options, None, token.clone(), None)
+    let telemetry_provider = Arc::new(documentdb_gateway::open_telemetry_provider::OpenTelemetryProvider::new());
+    
+    let telemetry_clone = telemetry_provider.clone();
```
AndrewKhoma, line 90: let's use `Arc::clone` instead of postfix notation to show that it's not a heavy-weight clone, but rather a rc increase

pg_documentdb_gw/src/main.rs
```diff
@@ -85,7 +85,21 @@ async fn main() {
     .await
     .unwrap();
 
-    run_server(service_context, certificate_options, None, token.clone(), None)
+    let telemetry_provider = Arc::new(documentdb_gateway::open_telemetry_provider::OpenTelemetryProvider::new());
```
AndrewKhoma, line 88: why is it Arc in the first place? i thought telemetry provider is owned by service context or my mental model is off?

pg_documentdb_gw/src/monitoring/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+
+use std::sync::Arc;
+use std::time::Duration;
+use tokio::time::interval;
+use log::{info, warn, error};
+
+use crate::postgres::Pool;
+use crate::telemetry::TelemetryProvider;
+
+/// ClusterMonitor periodically checks the availability of primary and secondary clusters
```
AndrewKhoma, line 17: can you specify which primary and secondary clusters are we getting connected to? in the current layout we only have 1 postgres instance per 1 instance of gw on the host machine, so I'm confused to see the secondary notation here

pg_documentdb_gw/src/monitoring/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+/// and emits metrics via the TelemetryProvider.
+pub struct ClusterMonitor {
+    primary_pool: Arc<Pool>,
+    secondary_pool: Option<Arc<Pool>>,
+    telemetry: Arc<dyn TelemetryProvider>,
```
AndrewKhoma, line 22: any chance of have it as generic instead of trait object? since we anyways know the concrete type of telemetry provider at from the main.rs at the moment of compilation

pg_documentdb_gw/src/monitoring/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+        if let Some(secondary_pool) = &self.secondary_pool {
+            let secondary_available = self.check_pool_availability(secondary_pool, "secondary").await;
+            self.telemetry.update_cluster_availability("secondary", secondary_available);
+        }
+    }
+
+    /// Check if a specific connection pool is available by attempting to get a connection
+    async fn check_pool_availability(&self, pool: &Pool, cluster_name: &str) -> bool {
+        match pool.get().await {
+            Ok(mut client) => {
```
AndrewKhoma, line 69: let's not name a pg connection a client, let's name it client instead, since we'll have an actual data client soon

pg_documentdb_gw/src/open_telemetry_provider.rs
```diff
@@ -0,0 +1,116 @@
...
+pub struct OpenTelemetryProvider {
+    traffic_counter: opentelemetry::metrics::Counter<u64>,
+    availability_gauge: opentelemetry::metrics::Gauge<i64>,
+}
+
+impl OpenTelemetryProvider {
+    pub fn new() -> Self {
+        // Initialize the global meter provider if not already done
+        let meter_provider = METER_PROVIDER.get_or_init(|| {
+            let endpoint = env::var("OTEL_EXPORTER_OTLP_ENDPOINT")
```
AndrewKhoma, line 39: can we capture this env variable through the config file?

pg_documentdb_gw/src/open_telemetry_provider.rs
```diff
@@ -0,0 +1,116 @@
...
+use std::sync::Arc;
+use std::time::Duration;
+use once_cell::sync::OnceCell;
+
+// OpenTelemetry imports
+use opentelemetry::{KeyValue, metrics::MeterProvider};
+use opentelemetry_otlp::{WithExportConfig};
+
+// Global meter provider for OpenTelemetry
+static METER_PROVIDER: OnceCell<Arc<opentelemetry_sdk::metrics::SdkMeterProvider>> = OnceCell::new();
```
AndrewKhoma, line 27: what are we trying to achieve here with Arc? since this smart pointer is basically the owner of this type and we're making the only instance of it as a singleton, then what's the logic behind ref counting in async context for this one?

pg_documentdb_gw/src/open_telemetry_provider.rs
```diff
@@ -0,0 +1,116 @@
...
+                .build();
+
+            Arc::new(provider)
+        });
+
+        let meter = meter_provider.meter("documentdb_gateway");
+
+        let traffic_counter = meter
+            .u64_counter("docdb_gateway_request_routing")
+            .with_description("Counts requests routed to primary or secondary regions")
```
AndrewKhoma, line 60: but gw is not in charge of regional failover, right? so maybe lets use this meter to just get the number of requests that gw served?

pg_documentdb_gw/src/open_telemetry_provider.rs
AndrewKhoma: and let's not use OnceCell, if it should be a single-owned, then let's move it to the ServiceContext and use it from there. I'm not a fan of this library and the way it lets you redefine the value it holds

AndrewKhoma (changes requested):
can we sync offline on what are we trying to achieve with the otel for gw monitoring, please?

PR #244, opened 2025-07-07
Add a public design for RBAC

lionelc:
This RBAC design doc presents a comprehensive framework that we plan to implement. It puts emphasis on the spec of role mapping, while the scope also includes policy-based integration.


nitinahuja89 (commented):
Here are my thoughts on the design.
1. To implement createRole and createUser privileges (and others that don't map to PG privileges) we can add authorization in the respective UDF's instead of relying on PG. The authorization check should retrieve the privileges for the user and the required privileges for the command from the catalog tables and determine eligibility to run the command. We should at least demonstrate this technique for a couple of such privileges so that we know that it works and can be easily leveraged in the future to support other privileges. 
2. Is the cost to move databases to separate schemas high? If we do this as a prerequisite we can allow database scoped authorization which is a important capability.
3. How are we handling insert on empty collection that requires inserting into the catalog table?
4. For documentdb_readwrite_role instead of blocking commands we should allowlist commands that it has the privileges to execute. Generally for authorization it is preferred to use allowlist instead of blocklist because adding new commands in the future would automatically allow access if we do nothing so it's not future safe.

lionelc:
> Here are my thoughts on the design.
> 
> 1. To implement createRole and createUser privileges (and others that don't map to PG privileges) we can add authorization in the respective UDF's instead of relying on PG. The authorization check should retrieve the privileges for the user and the required privileges for the command from the catalog tables and determine eligibility to run the command. We should at least demonstrate this technique for a couple of such privileges so that we know that it works and can be easily leveraged in the future to support other privileges.
> 2. Is the cost to move databases to separate schemas high? If we do this as a prerequisite we can allow database scoped authorization which is a important capability.
> 3. How are we handling insert on empty collection that requires inserting into the catalog table?
> 4. For documentdb_readwrite_role instead of blocking commands we should allowlist commands that it has the privileges to execute. Generally for authorization it is preferred to use allowlist instead of blocklist because adding new commands in the future would automatically allow access if we do nothing so it's not future safe.

Sorry for the late reply @nitinahuja89 . 

1. The design comes with the assumption that we just use PG-compatible privileges and that would be adequate. And UDF as a role wrap-up is not preferrable in security context because the built-in roles has built-in PG security too, while adding a UDF could mean adding vulnerability to it. So, we would need to discuss on a good solution if there are more privileges that need to be supported.

2. Yes, the cost of migrating DBs is very high.

3. In this case, it's effectively a collection create. We will only allow INSERT privileges on an already-existing collection so this situation will not arise in our case.

4. The allow-list based implementation would bring a lot of breaking changes for the built-in structures. That can be planned at a later phase, and we should evaluate the feasibility carefully.

nitinahuja89:
> > Here are my thoughts on the design.
> > 
> > 1. To implement createRole and createUser privileges (and others that don't map to PG privileges) we can add authorization in the respective UDF's instead of relying on PG. The authorization check should retrieve the privileges for the user and the required privileges for the command from the catalog tables and determine eligibility to run the command. We should at least demonstrate this technique for a couple of such privileges so that we know that it works and can be easily leveraged in the future to support other privileges.
> > 2. Is the cost to move databases to separate schemas high? If we do this as a prerequisite we can allow database scoped authorization which is a important capability.
> > 3. How are we handling insert on empty collection that requires inserting into the catalog table?
> > 4. For documentdb_readwrite_role instead of blocking commands we should allowlist commands that it has the privileges to execute. Generally for authorization it is preferred to use allowlist instead of blocklist because adding new commands in the future would automatically allow access if we do nothing so it's not future safe.
> 
> Sorry for the late reply @nitinahuja89 .
> 
> 1. The design comes with the assumption that we just use PG-compatible privileges and that would be adequate. And UDF as a role wrap-up is not preferrable in security context because the built-in roles has built-in PG security too, while adding a UDF could mean adding vulnerability to it. So, we would need to discuss on a good solution if there are more privileges that need to be supported.
> 2. Yes, the cost of migrating DBs is very high.
> 3. In this case, it's effectively a collection create. We will only allow INSERT privileges on an already-existing collection so this situation will not arise in our case.
> 4. The allow-list based implementation would bring a lot of breaking changes for the built-in structures. That can be planned at a later phase, and we should evaluate the feasibility carefully.

We do need to support a lot of privileges that are not natively supported in PG so we need a concrete plan for that. 
What is the breaking changes for allow-list based implementation? Doing this in phase 1 is the best because it sets the stage  for the scheme we want to follow and all new privilege additions should go through this route of being allowlisted.

xgerman:
> Let's say the gateway is configured to connect to PostgreSQL using the `postgres` user. User `alex` connects to the gateway. What happens next? What credentials are used to connect to PostgreSQL? Where is authentication and authorization happening? After reading the document multiple times, I'm still unable to answer those questions.

Our current gateway will maintain a connection pool per user so you will login as alex. Ina uture iteration we migth add a way to use a *sing;e* collection pool and wire up the functions so theyc an run as a specific user and authrization will happen during said context switch.

xgerman:
closing - superseded by other PRs

PR #262, opened 2025-08-06
Migrating functions to new schemas to support RBAC

DeepakSPemmaraju:
This is a design doc about how we're migrating our functions to new schemas in order to support RBAC

### DIFF
--- a/docs/v1/rbac/schema-changes-for-rbac.md
+++ b/docs/v1/rbac/schema-changes-for-rbac.md
@@ -1,0 +1,472 @@
+---
+
+uid: schema_changes_for_rbac.md
+title: Schema changes to facilitate RBAC in DocumentDB
+
+---
+
+# Changing Default Permissions of Functions in DocumentDB
+
+Engineering Owner: deepakpe
+
+## Overview
+
+Today the execution privileges on all the functions in our schemas are set to PUBLIC.
+In light of the upcoming feature to support RoleBasedAccessControl we want to remove the ability to execute functions from public from all our schemas and grant access as needed to specific built-in roles.
+
+## Design
+
+Today we have 4 built-in roles in documentdb.
+
+1. documentdb role : This role is a superuser. As a result it can execute all functions in all the schemas without us having to grant it any specific privileges.
+2. documentdb_admin_role : This role can perform read write operations across the entire cluster. In addition it can also perform admin operations such as sharding. This is the role users of documentdb_local get by default.
+3. documentdb_readonly_role: This role has read access across the entire cluster.
+4. documentdb_bg_worker_role: This is role is used to run background processes.
+
+The functions in documentdb are spread across three schemas.
+
+1. documentdb_api
+2. documentdb_api_internal
+3. documentdb_api_catalog
+
+### documentdb_api
+
+The table below lists the current functions and procedures in documentdb_api and which built-in roles are allowed to execute them. 
+We will not be changing the documentdb_api schema itself.
+Instead we will create a new schema documentdb_api_v2 where the ability to execute functions is revoked from PUBLIC and the ability to execute specific functions is granted only to specific built-in roles.
+
+| Schema         | Name                               | Type | Built-in roles that have execute privileges                           |
+|----------------|------------------------------------|------|-----------------------------------------------------------------------|
+| documentdb_api | create_user                        | func | documentdb                                                            |
+| documentdb_api | drop_user                          | func | documentdb                                                            |
+| documentdb_api | update_user                        | func | documentdb (but every user can use this to change their own password) |
+| documentdb_api | binary_extended_version            | func | documentdb_admin_role                                                 |
+| documentdb_api | binary_version                     | func | documentdb_admin_role                                                 |
+| documentdb_api | coll_mod                           | func | documentdb_admin_role                                                 |
+| documentdb_api | collection                         | func | documentdb_admin_role                                                 |
+| documentdb_api | compact                            | func | documentdb_admin_role                                                 |
+| documentdb_api | create_collection                  | func | documentdb_admin_role                                                 |
+| documentdb_api | create_collection_view             | func | documentdb_admin_role                                                 |
+| documentdb_api | create_indexes_background          | func | documentdb_admin_role                                                 |
+| documentdb_api | delete                             | func | documentdb_admin_role                                                 |
+| documentdb_api | drop_collection                    | func | documentdb_admin_role                                                 |
+| documentdb_api | drop_database                      | func | documentdb_admin_role                                                 |
+| documentdb_api | drop_indexes                       | proc | documentdb_admin_role                                                 |
+| documentdb_api | insert                             | func | documentdb_admin_role                                                 |
+| documentdb_api | insert_bulk                        | proc | documentdb_admin_role                                                 |
+| documentdb_api | insert_one                         | func | documentdb_admin_role                                                 |
+| documentdb_api | rename_collection                  | func | documentdb_admin_role                                                 |
+| documentdb_api | reshard_collection                 | func | documentdb_admin_role                                                 |
+| documentdb_api | shard_collection                   | func | documentdb_admin_role                                                 |
+| documentdb_api | shard_collection                   | func | documentdb_admin_role                                                 |
+| documentdb_api | unshard_collection                 | func | documentdb_admin_role                                                 |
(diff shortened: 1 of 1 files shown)


docs/v1/rbac/schema-changes-for-rbac.md
```diff
@@ -0,0 +1,472 @@
...
+## Overview
+
+Today the execution privileges on all the functions in our schemas are set to PUBLIC.
+In light of the upcoming feature to support RoleBasedAccessControl we want to remove the ability to execute functions from public from all our schemas and grant access as needed to specific built-in roles.
+
+## Design
+
+Today we have 4 built-in roles in documentdb.
+
+1. documentdb role : This role is a superuser. As a result it can execute all functions in all the schemas without us having to grant it any specific privileges.
```
nitinahuja89, line 21: can only documentdb role today create new roles?

docs/v1/rbac/schema-changes-for-rbac.md
```diff
@@ -0,0 +1,472 @@
...
+2. documentdb_admin_role : This role can perform read write operations across the entire cluster. In addition it can also perform admin operations such as sharding. This is the role users of documentdb_local get by default.
```
nitinahuja89, line 22: which functions can this role not execute today?

docs/v1/rbac/schema-changes-for-rbac.md
```diff
@@ -0,0 +1,472 @@
...
+
+1. documentdb_api
+2. documentdb_api_internal
+3. documentdb_api_catalog
+
+### documentdb_api
+
+The table below lists the current functions and procedures in documentdb_api and which built-in roles are allowed to execute them. 
+We will not be changing the documentdb_api schema itself.
+Instead we will create a new schema documentdb_api_v2 where the ability to execute functions is revoked from PUBLIC and the ability to execute specific functions is granted only to specific built-in roles.
```
nitinahuja89, line 41: can the functions from v1 schema still be used?

docs/v1/rbac/schema-changes-for-rbac.md
```diff
@@ -0,0 +1,472 @@
...
+
+| Schema         | Name                               | Type | Built-in roles that have execute privileges                           |
+|----------------|------------------------------------|------|-----------------------------------------------------------------------|
+| documentdb_api | create_user                        | func | documentdb                                                            |
+| documentdb_api | drop_user                          | func | documentdb                                                            |
+| documentdb_api | update_user                        | func | documentdb (but every user can use this to change their own password) |
```
nitinahuja89: do we have checks to ensure only the password is changed and no other attributes are checked to prevent privilege escalation?

docs/v1/rbac/schema-changes-for-rbac.md
```diff
@@ -0,0 +1,472 @@
...
+| documentdb_api | users_info                         | func | documentdb_admin_role, documentdb_readonly_role                       |
+| documentdb_api | validate                           | func | documentdb_admin_role, documentdb_readonly_role                       |
+
+### documentdb_api_catalog
+
+All the functions and procedures in documentdb_api_catalog are read only. As a result both documentdb_admin_role and documentdb_readonly_role have execute privileges on all functions and procs in documentdb_api_catalog. As a result we will not be making any changes to this schema.
+
+### documentdb_api_internal
+
+As part of this work we want to divide the documentdb_api_internal schema into four seperate schemas namely
```
nitinahuja89, line 92: This is good for coarse grained privileges but doesn't help with fine grained privilege control. For example a custom role might have been granted insert privilege but not update privilege. That role needs execute privilege on insert_one but not on update_one.

docs/v1/rbac/schema-changes-for-rbac.md
```diff
@@ -0,0 +1,472 @@
...
+          ApiInternalReadWriteSchemaName = "documentdb_api_internal_readwrite";
+          ApiInternalAdminSchemaName = "documentdb_api_internal_admin";
+          ApiInternalBgWorkerSchemaName = "documentdb_api_internal_bgworker";
+      }
+  }
+  ```
+
+### Gateway Changes
+
+All calls from the gateway to the backend go through command_funcs.c which in turn calls the methods in oid_cache.c. We will therefore make sure gateway switches over to the new schema by replacing the schema name variables used in oid_cache.c.
```
nitinahuja89, line 432: Will this also be controlled through a configuration?

docs/v1/rbac/schema-changes-for-rbac.md
```diff
@@ -0,0 +1,472 @@
...
+
+We need to rollout this feature over several releases.
+
+Release 1:
+
+- Create the new schemas and add functions to the new schemas
+- Introduce new global variables ApiSchemaNameV3, ApiInternalReadOnlySchemaName, ApiInternalReadWriteSchemaName, ApiInternalAdminSchemaName, ApiInternalBgWorkerSchemaName and replace old schema variables with the new variables. The new variables will continue to have the old schema names by default.
+- Introduce the new GUCs and code in PGInit to override the schema variable names when the GUCs are set
+
+Release 2:
```
nitinahuja89, line 444: What is this release doing? Looks like these are just tests. The actual switch of the GUC happens in release 3.

docs/v1/rbac/schema-changes-for-rbac.md
```diff
@@ -0,0 +1,472 @@
...
+
+  ``` sql
+  CREATE OR REPLACE FUNCTION documentdb_api_v2.create_user(p_spec __CORE_SCHEMA_V2__.bson)
+                                                            RETURNS __CORE_SCHEMA_V2__.bson
+  LANGUAGE C
+  VOLATILE
+  AS 'MODULE_PATHNAME', $function$documentdb_extension_create_user$function$;
+  ```
+
+- GRANT EXECUTE on the functions in these schemas to the respective built-in roles.
```
nitinahuja89, line 381: Are we granting execute at the function level or schema level?

docs/v1/rbac/schema-changes-for-rbac.md
DeepakSPemmaraju: @nitinahuja89  Yes

docs/v1/rbac/schema-changes-for-rbac.md
DeepakSPemmaraju: @nitinahuja89 Today the documentdb_admin_role cannot call user Create/Update/Drop functions. It can call everything else

docs/v1/rbac/schema-changes-for-rbac.md
DeepakSPemmaraju: Until we fully transition to V2, then we plan to deprecate them

docs/v1/rbac/schema-changes-for-rbac.md
DeepakSPemmaraju: @nitinahuja89  PG takes care of this for us. It ensures that a user without the CreateRole privilege can only update their own password and nothing else. And today we don't explicitly grant CreateRole to any other roles

docs/v1/rbac/schema-changes-for-rbac.md
DeepakSPemmaraju: Agreed, this is only to make maintaining these schemas more manageable. It cannot replace us having to grant execute selectively when creating custom roles

docs/v1/rbac/schema-changes-for-rbac.md
DeepakSPemmaraju: Yes, we're controlling the switch over with a GUC

docs/v1/rbac/schema-changes-for-rbac.md
DeepakSPemmaraju: Release 1 adds the new shcemas and functions, release 2 adds tests and the ability to optin a few test clusters, release 3 makes the new schemas default and release 4 cuts off access to the old schemas

nitinahuja89: Doesn't look like release 2 is actually releasing any capability. This could just be rolled into release 1.

docs/v1/rbac/schema-changes-for-rbac.md
nitinahuja89: I believe you mentioned in the above comment that we will grant execute privileges for individual functions instead of granting execute on the entire schema. Is that correct?

PR #284, opened 2025-08-25
Add JsonB op class support to documentdb_rum

visridha:
Add simple support as a reference implementation sketch of jsonbpath with simple comparison operators for documentdb_rum.

Note: this needs to be extended significantly to support more scenarios but presents a sample of what can be done.


visridha (commented):
Waiting to prevent completion

internal/pg_documentdb_extended_rum/src/test/regress/expected/basic_jsonb_query_pushdown.out
```diff
@@ -0,0 +1,113 @@
...
+-----------------------------------------------------------------------------------------------------------
+ Index Scan using my_idx_ab on jsonb_basic_query
+   Index Cond: (val OPERATOR(documentdb_extended_rum_jsonb_ops_internal.#@@) '($."a"."b" == 2)'::jsonpath)
+   Filter: (val @@@ '($."a"."b" == 2)'::jsonpath)
+(3 rows)
+
+EXPLAIN (COSTS OFF) SELECT * FROM public.jsonb_basic_query WHERE val @@@ '$.a.b.c == 2';
+                      QUERY PLAN                      
+------------------------------------------------------
+ Seq Scan on jsonb_basic_query
```
safern, line 49: shouldn't this have been an index scan given the a.b being a wildcard index?

PR #293, opened 2025-08-27
Proposes a refactoring of the crates into a more resuable strcuture

xgerman:
We propose to refactor the crate strcuture to aid in resuability and maintenance.


cooltodinesh:
It's convenient if stuff that should build together, stays together. If there are more than one possibilities of implementations, we can think of new crate.
- IMO mongo protocol handling still could be part of core crate. We would need a Crate for all trait's (interfaces) which core gw will call and other crates would implement for stuff like metric providers, authentication providers.
- macros -> could it be module instead of Crate? We would need it while building core gw. If building macros Crate always need rebuild of core-gw crate, they might as well stay together.

Do you think we can start with folders/modules for now and move to Crate for specific folders where we fill it could be provided as one of implementations in runtime?

cooltodinesh:
1. I see, but having mongo protocol outside of core Crate seems bit odd. If we put trait and default protocol implementation in core crate itself, you could override implementation of that trait in internal branch with another custom crate, would it work?
2. Agree, them in core crate seems fine for now.

AndrewKhoma:
Hey @cooltodinesh!

> 1. I see, but having mongo protocol outside of core Crate seems bit odd. If we put trait and default protocol implementation in core crate itself, you could override implementation of that trait in internal branch with another custom crate, would it work?

Yes, basically the overall idea is the following - other teams wants to build on top of the gateway and as part of being good citizens we're doing a split on core functionality which can be reused by other vs mongo specific. So the top crate is still remains mongo-specific, but if someone wants to built their own connection pooling or other stuff - feel free to do so.

> Do you think we can start with folders/modules for now and move to Crate for specific folders where we fill it could be provided as one of implementations in runtime?

yeah, that PR is coming, hang tight :)

docs/v1/crate_structure_design.md
```diff
@@ -0,0 +1,24 @@
...
+
+* `oss-docdb-gw`
+  * Implements MongoDB wire-protocol handling: request parsing, authentication, routing to processing logic, and response construction. Depends on the core crate for database access and observability.
+
+* `macros`
+  * A small procedural-macro crate containing code-generation helpers to reduce boilerplate across the gateway.
+
+Benefits:
+
+* Clear separation of concerns — protocol handling, storage, and cross-cutting services are separated, improving modularity and enabling reuse of the storage core across multiple protocol gateways.
```
AndrewKhoma, line 20: we still don't have storage, I would drop it from design

xgerman:
@tsharp will take over

tsharp:
@xgerman - I am closing this PR for now as this work looks largely complete even though the directories / crates have different names.

I'll take some time to document this as more of a statement in `pg_documentdb_gw/README.md`

Our current crates are:
- [documentdb_gateway](https://github.com/documentdb/documentdb/tree/main/pg_documentdb_gw/documentdb_gateway)
- [documentdb_gateway_core](https://github.com/documentdb/documentdb/tree/main/pg_documentdb_gw/documentdb_gateway_core)
- [documentdb_macros](https://github.com/documentdb/documentdb/tree/main/pg_documentdb_gw/documentdb_macros)

We'll also have a new crate upcoming for traits that is `documentdb_gateway_lib` as well and enforce crate reference direction / boundaries.

PR #303, opened 2025-09-04
Add similarity check

fatmelon:
Add similarity check


.github/workflows/similarity-analysis.yml
sivethe, line 1: Please pickup the latest version from https://github.com/sivethe/similaritychecker/blob/v2/.github/workflows/similarity-analysis.yml.  We need to parameterize the source and target repos

scripts/extract_strings_from_cplusplus.py
sivethe, line 1: How about creating a tools/similarityChecker folder where we can put the scripts, README and the test cases from the original repo?

sivethe (changes requested):

.github/workflows/similarity-analysis.yml
fatmelon: I think the source repo is always documentdb, right? Then we only need to parameterize a target repo

sivethe: We want to enable contributors who have forked this repo to also be able to run the pipeline in their forks.  We can keep the source as the current repo

AndrewKhoma: to build on top of what Siddhesh proposed: how about reusable workflow?
https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows

so we have a template for similarity analysis and another one is dispatching the workflow calls

fatmelon: the latest commit uses current repo as the source repo and allow contributors manually parameterize a target repo.
@AndrewKhoma could you share more details about the two workflow you mentioned :)?  What are the advantages of this design compared with use single GH action ?

AndrewKhoma: so take my suggestion with the grain of salt here, since I've used GH actions like 5 years ago.

what I'm saying is maybe we can move all of the logic in the template (i've looked it up and says that it's called reusable workflow on GH) and then the caller logic to the separate pipeline that will be just passing values to it. from my standpoint it will benefit others by a better decoupling of caller values from the actual pipeline logic in template, but once again, feel free to drop it

xgerman:
We don't need this change any longer.

xgerman:
We don't need  this any longer

PR #305, opened 2025-09-05
DocumentDB sync 20250905

### LINKED ISSUE #295: Concurrent upserts lead to data loss, that is not being observed in MongoDB
**Describe the bug**

DocumentDB fails to properly handle concurrent upsert operations on the same document, causing one operation to overwrite the other and resulting in data loss. 

This race condition does not occur in standard MongoDB.

**Environment**
- DocumentDB version: ghcr.io/microsoft/documentdb/documentdb-local:latest (17198796944-2025-08-25-amd64)
- MongoDB: 6, 7, 8
- Architecture: Docker containerized local instance

**Reproduction Steps**

Prerequisites
Repository: https://github.com/gliwka/documentdb-race-reproducer

```bash
git clone https://github.com/gliwka/documentdb-race-reproducer
cd documentdb-race-reproducer
npm install
npm run docker:up
npm run test-documentdb
```

Queries that reproduce the issue:
```typescript
// Two concurrent upserts on same document (on different connections)
await Promise.all([
  collection1.updateOne(
    { _id: userId },
    { $unset: { deleted: "" }, $set: { isProxy: false } },
    { upsert: true }
  ),
  collection2.updateOne(
    { _id: userId },
    { $addToSet: { profileIds: profileId } },
    { upsert: true }
  )
]);
```

**Expected behavior**

Both upsert operations should be applied atomically to the same document, resulting in:
```json
{
  "_id": "test-id",
  "isProxy": false,
  "profileIds": ["some-profile-id"]
}
```

**Actual behavior**

One operation overwrites the other, causing partial data loss:
- 98% data loss rate (only 2/100 iterations successful)
- Documents missing either `isProxy` field or `profileIds` array
- MongoDB control test shows 0% data loss (100/100 successful)

**Additional context**

We stumbled onto this issue while transferring our workloads from MongoDB to the CosmosDB vCore offering. Seems like the behaviour on data races is not compatible. 

With transactions, both behave the same. But without, MongoDB applies the upserts as expected. DocumentDB leads to silent data loss

shuaitian-git:
resolves #295


CHANGELOG.md
```diff
@@ -1,10 +1,13 @@
...
+* Top-level `let` variables and `$$NOW` supported by default.
+* Fix collation support on find and aggregation when variableSpec is not available.
+* Support `dropRole` command *[Feature]*
 
-### documentdb v1.107-0 (Unreleased) ###
-- Support sort by _id against the _id index using the enableIndexOrderbyPushdown flag *[Feature]*.
-- Improvements to explain for various scan types *[Feature]*.
+### documentdb v0.107-0 (Unreleased) ###
+* Support sort by _id against the _id index using the enableIndexOrderbyPushdown flag *[Feature]*.
+* Improvements to explain for various scan types *[Feature]*.
```
lichoil, line 8: why did it exclude below feature in 1.107?
- [Preview] Support schema enforcement with CSFLE integration
- [Preview] Validate $jsonSchema syntax during rule creation or modification (schema validation)

CHANGELOG.md
```diff
@@ -1,10 +1,13 @@
-### documentdb v1.108-0 (Unreleased) ###
+### documentdb v0.108-0 (Unreleased) ###
+* Top-level `let` variables and `$$NOW` supported by default.
+* Fix collation support on find and aggregation when variableSpec is not available.
+* Support `dropRole` command *[Feature]*
 
-### documentdb v1.107-0 (Unreleased) ###
-- Support sort by _id against the _id index using the enableIndexOrderbyPushdown flag *[Feature]*.
-- Improvements to explain for various scan types *[Feature]*.
+### documentdb v0.107-0 (Unreleased) ###
```
AndrewKhoma, line 6: nit: can we add that gw now have the following:
1. update the logging levels
2. provided request context struct to improve the code readability and reusage
3. rework the TLS layer

AndrewKhoma: i feel like we should update the changelog from our side for each of the PRs so it will ease this process

CHANGELOG.md
shuaitian-git: I guess the author forgot to add those here, we can fix this internally

CHANGELOG.md
shuaitian-git: for adding gw features, let's do it separately and out of this sync PR.

And yes, we're working on a plan to move all documentdb code changes to GitHub, with that we don't need this sync anymore and we can maintain this changelog here.

AndrewKhoma: ack. thanks for providing guidance!

lichoil (approved):

PR #310, opened 2025-09-09
Package: update changelog/spec when building packages

shuaitian-git:


packaging/deb/packaging-entrypoint.sh
```diff
@@ -6,6 +6,19 @@ test -n "$OS" || (echo "OS not set" && false)
 # Change to the build directory
 cd /build
 
+# Update packaging changelogs from CHANGELOG.md (fail build if this fails)
```
fatmelon, line 9: what if this part fails to get `DOCUMENTDB_VERSION`? will the build break somewhere?

AndrewKhoma: valid, I would propose to add `exit 1` after the echo WARNING

visridha (approved):

PR #314, opened 2025-09-15
[Do NOT MERGE] Adds a documentdb gateway test gate

xgerman:
Adds a gate to test the gateway


.github/workflows/gw_tests.yml
```diff
@@ -0,0 +1,154 @@
...
+          libxml2-dev \
+          libjson-c-dev \
+          libgdal-dev \
+          libgeos++-dev \
+          libgeos-dev \
+
+          cp ./scripts/install_setup_postgis.sh /tmp/install_setup/
+          sudo INSTALL_DEPENDENCIES_ROOT=$INSTALL_DEPENDENCIES_ROOT PGVERSION=$PG_VERSION /tmp/install_setup/install_setup_postgis.sh
+
+          curl https://sh.rustup.rs -sSf | sh -s -- -y --no-modify-path --default-toolchain stable
```
jcagme: Can you try fetching 1.86.0?

.github/workflows/gw_tests.yml
```diff
@@ -0,0 +1,154 @@
...
+          - arch: amd64
+            runner: ubuntu-22.04
+          - arch: arm64
+            runner: ubuntu-22.04-arm
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      
+      - name: Install package dependencies
+        run: |
```
AndrewKhoma, line 48: we share this script with the regress_tests.yml pipeline. Can we not duplicate ourselves so it will be either to maintain

.github/workflows/gw_tests.yml
AndrewKhoma: and i would suggest to have it as a separate step

.github/workflows/gw_tests.yml
```diff
@@ -0,0 +1,154 @@
...
+          curl https://sh.rustup.rs -sSf | sh -s -- -y --no-modify-path --default-toolchain stable
+          rustc --version && cargo --version
+     
+      - name: Export pg_config PATH
+        run: |
+          echo "/usr/lib/postgresql/${{ matrix.pg_version }}/bin" >> $GITHUB_PATH
+
+      - name: Clippy Check
+        run: |
+          cd pg_documentdb_gw
```
AndrewKhoma: i've found out there's a working-directory keyword: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepsworking-directory

.github/workflows/gw_tests.yml
```diff
@@ -0,0 +1,154 @@
...
+    paths-ignore:
+      - 'docs/**'
+      - '.devcontainer/**'
+      - '*.md'
+  pull_request:
+    types: [opened, reopened, synchronize, ready_for_review]
+    paths-ignore:
+      - 'docs/**'
+      - '.devcontainer/**'
+      - '*.md'
```
AndrewKhoma, line 19: maybe instead we can do:
```
paths:
  - 'pg_documentdb_gw/**'
```

.github/workflows/gw_tests.yml
jcagme: Agreed

.github/workflows/gw_tests.yml
xgerman: I am on the fence for this one. How would we protect against an engine change which break the gw? I think we would want to run the integration test with any engine change as well...

AndrewKhoma: nice thought! agree

AndrewKhoma: hmm, then let's apparently just modify the same pipeline as for the engine to just build gateway on top and run tests then?

.github/workflows/gateway_tests.yml
```diff
@@ -0,0 +1,166 @@
...
+          which pg_config
+          make
+          sudo PATH=$PATH make install
+
+      # - name: Clippy Check
+      #   run: |
+      #     cargo clippy --all-targets --all-features -- -D warnings
+      #   working-directory: pg_documentdb_gw
+
+      - name: Gateway Integration Test
```
jcagme: Looks like this step starts DocumentDB but doesn't run tests?

xgerman: yes, tests are in the next step

xgerman:
Looks like my work got replicated in https://github.com/documentdb/documentdb/pull/351 - so closing.

PR #316, opened 2025-09-16
Adding a github workflow to run mongo's compatibility tests

vtarassov:
Adding a workflow that downloads and runs Mongo's compatibility tests:
https://github.com/mongodb-developer/service-tests/tree/master/Compatibility%20Websites%20Code

This is mostly meant as a stop-gap until we develop a full suite of our own compatibility tests, but in the meantime it can serve as a good way to exercise a wide selection of end-to-end use-cases against the gateway and engine. 

I have only tested this locally using [nektos/act](https://github.com/nektos/act) and can't speak for how well it'll behave in github. 

```
act -j mongodb-compat-tests-ubuntu --matrix arch:amd64 --matrix pg_version:16
```

### DIFF
--- a/.github/workflows/mongodb-compat-tests.yml
+++ b/.github/workflows/mongodb-compat-tests.yml
@@ -1,0 +1,145 @@
+name: MongoDB Compatibility Tests 
+
+# Ensure we don't run duplicate workflows on the same branch/PR
+concurrency:
+  group: mongodb-compat-tests-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+on:
+  workflow_dispatch:
+  pull_request_review:
+    types: [submitted]   
+
+jobs:
+  mongodb-compat-tests-ubuntu:
+    if: github.event.review.state == 'approved' || github.actor == 'nektos/act'
+    runs-on: ${{ matrix.runner }}
+    name: Run mongo compatibility tests on ${{ matrix.runner }} with PG ${{ matrix.pg_version }}
+    strategy:
+      fail-fast: false
+      matrix:
+        pg_version:
+          - 16
+          - 17
+        arch: [amd64, arm64]
+        include:
+          - arch: amd64
+            base_arch: AMD64
+            runner: ubuntu-22.04
+          - arch: arm64
+            base_arch: ARM64
+            runner: ubuntu-22.04-arm
+
+    env:
+      DEFAULT_BASE_VERSION: '0.104.0'
+      IMAGE_NAME: documentdb-local
+      IMAGE_TAG: ${{ github.run_id }}-$(date +'%Y-%m-%d')
+      TEST_USER: test_user
+      DOCDB_PORT: 10260
+      PG_PORT: 9719
+      
+    steps:
+      - name: Checkout DocumentDB repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Install required dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y build-essential curl python3-pip
+          python3 -m pip install pymongo
+
+      - name: Set up environment
+        run: |
+          echo "Setting up test environment"
+          mkdir -p test-results
+          echo "BASE_VERSION=${{ github.event.inputs.base_version || env.DEFAULT_BASE_VERSION }}" >> $GITHUB_ENV
+          
+          # Generate a random password for this test run. 
+          TEST_PWD=$(openssl rand -base64 32 | tr -d '/+=\n')
+          echo "TEST_PWD=$TEST_PWD" >> $GITHUB_ENV
+          
+          # Set up MongoDB connection URL with the generated password
+          MONGO_URL="mongodb://$TEST_USER:$TEST_PWD@localhost:$DOCDB_PORT/?authMechanism=SCRAM-SHA-256&tls=true&tlsAllowInvalidCertificates=true&ssl=true"
+          echo "MONGO_URL=$MONGO_URL" >> $GITHUB_ENV
+
+      - name: Build and start DocumentDB
+        id: start-documentdb
+        run: |
+          echo "Building image"
+          # Build Docker image for DocumentDB
+          TAG=${{ env.IMAGE_TAG }}-${{ matrix.arch }}
+          docker build \
+            --build-arg BASE_ARCH=${{ matrix.base_arch }} \
+            --build-arg BASE_VERSION=${{ env.BASE_VERSION }} \
+            --build-arg PG_VERSION=${{ matrix.pg_version }} \
+            -t ghcr.io/${{ github.repository }}/${{ env.IMAGE_NAME }}:$TAG \
+            -f .github/containers/Build-Ubuntu/Dockerfile_gateway .
+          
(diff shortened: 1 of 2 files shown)


AndrewKhoma:
Hey @vtarassov! thanks for your changes, i need some CELA guidance here, since the LICENSE here:
https://github.com/mongodb-developer/service-tests/blob/master/LICENSE
is SSPL even though it was updated last time 4 years ago. @visridha are we good here?

.github/workflows/mongodb-compat-tests.yml
```diff
@@ -0,0 +1,159 @@
...
+            -e DOCUMENTDB_PORT=$DOCDB_PORT \
+            -e POSTGRESQL_PORT=$PG_PORT \
+            ghcr.io/${{ github.repository }}/${{ env.IMAGE_NAME }}:$TAG
+          
+          echo "Waiting for DocumentDB server to be ready..."
+          max_retries=30
+          retry_count=0
+          
+          while [ $retry_count -lt $max_retries ]; do
+            if docker exec documentdb-server pg_isready -p $PG_PORT -h localhost; then
```
sivethe: This only checks for PG and not our GW.  How about using mongosh --build-info (https://www.mongodb.com/docs/mongodb-shell/reference/options/#std-option-mongosh.--build-info) to verify connectivity via mongo endpoint

.github/workflows/mongodb-compat-tests.yml
```diff
@@ -0,0 +1,159 @@
...
+          # Capture compatibility percentage in a variable
+          COMPATIBILITY_PERCENTAGE=$(grep "Compatibility Percentage" test_output.txt | grep -o '[0-9]\+\.[0-9]\+')
+
+          # TODO: Compare against baseline
+          
+          # Display the compatibility percentage
+          echo "Compatibility Percentage: $COMPATIBILITY_PERCENTAGE"
+         
+
+      - name: Upload Compatibility Report
```
sivethe: Given that the repo is SSPL licensed, I would skip the upload step completely to prevent any concerns around service-usage (https://en.wikipedia.org/wiki/Server_Side_Public_License#cite_note-1).  We should compare against baseline and add comments about which specific tests are failing.  The onus is on the committer to rerun these tests on their fork/private machine and verify that the tests are passing.

.github/workflows/mongodb-compat-tests.yml
```diff
@@ -0,0 +1,159 @@
...
+          - arch: arm64
+            base_arch: ARM64
+            runner: ubuntu-22.04-arm
+
+    env:
+      DEFAULT_BASE_VERSION: '0.104.0'
+      IMAGE_NAME: documentdb-local
+      IMAGE_TAG: ${{ github.run_id }}-$(date +'%Y-%m-%d')
+      TEST_USER: test_user
+      TEST_PWD: docdb123
```
sivethe: Can we use a random password instead of a fixed one?

sivethe (changes requested):

.github/workflows/mongodb-compat-tests.yml
```diff
@@ -0,0 +1,159 @@
...
+on:
+  push:
+    branches:
+      - 'main'
+    paths-ignore:
+      - 'docs/**'
+      - '.devcontainer/**'
+      - '*.md'
+  pull_request:
+    types: [opened, reopened, synchronize]
```
xgerman: I had a conversation with legal and we can only run that at merge. see https://stackoverflow.com/questions/60710209/trigger-github-actions-only-when-pr-is-merged on how to set that up.

sivethe: @xgerman, this is contrary to my understanding.  We need these pipelines to prevent regressions being checked-in so we should run them pre-checkin.  What we dont want to allow is the ability for anyone to trigger these workflows in Github (aka manual runs on developer branches).  We should only run these changes on active PRs on the main/release branch

vtarassov: Can we meet in the middle, running the workflow only once the PR is approved?
https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#running-a-workflow-when-a-pull-request-is-approved

It should then also be possible to configure branch protections to wait for this workflow to succeed before merging.

sivethe: That works too.

.github/workflows/mongodb-compat-tests.yml
```diff
@@ -0,0 +1,145 @@
...
+  cancel-in-progress: true
+
+on:
+  workflow_dispatch:
+  pull_request_review:
+    types: [submitted]   
+
+jobs:
+  mongodb-compat-tests-ubuntu:
+    if: github.event.review.state == 'approved' || github.actor == 'nektos/act'
```
sivethe, line 15: Please add a section in CONTRIBUTING.md on how to use act to run/validate the compatibility tests locally

.github/workflows/mongodb-compat-tests.yml
```diff
@@ -0,0 +1,145 @@
...
+          sed -i "s|RESULT_DB_URI = .*|RESULT_DB_URI = '$(echo "$MONGO_URL" | sed 's/&/\\&/g')'|" ./config.py
+          
+          # Run the tests
+          echo "Running Tests"
+          python3 run_test.py | tee test_output.txt
+          
+          # Capture compatibility percentage in a variable
+          COMPATIBILITY_PERCENTAGE=$(grep "Compatibility Percentage" test_output.txt | grep -o '[0-9]\+\.[0-9]\+')
+
+          # TODO: Compare against baseline
```
sivethe, line 137: Is this planned in a separate PR?

sivethe (changes requested):
The changes look good.  Primary question around baseline comparison and how we intend to fail the workflow if we detect regression in compat

.github/workflows/mongodb-compat-tests.yml
vtarassov: My hesitation here is that there's a major caveat with using this locally. It uses Dockerfile_gateway, which builds the GW from source but uses pre-built extension binaries. In order to be fully usable for development it should rebuild both from local source. 

I'll put together a pattern for this.

.github/workflows/mongodb-compat-tests.yml
vtarassov: Yeah. I wanted to take some time to find a good pattern for this since we're liable to repeat it, but wanted to make this test available in the meantime. I can roll that into this change as well if you prefer.

.github/workflows/mongodb-compat-tests.yml
AndrewKhoma: hmm, that's actually a very valid reason!

.github/workflows/mongodb-compat-tests.yml
sivethe: No makes sense.  Lets add it in another PR

.github/workflows/mongodb-compat-tests.yml
shuaitian-git: @vtarassov there is an active PR trying to update Dockerfile_gateway, which will expect an deb package as argument, then build GW from source code. With this change, we can build package first and then run this dockerfile in a workflow. See #331 

We also plan to include gw into packaging build, so this dockerfile will be simpler.

xgerman:
@vtarassov can we close since we want to achive the same via the functional tests

PR #317, opened 2025-09-16
[Gateway][Rust] Adding workerThreads parameter to control tokio runtime parallelism

vtarassov:
Under the current implementation, the tokio runtime defaults to one active thread. This became a bottleneck for me when trying to run north of 20,000 requests/sec. 

This change switches to using a multi-threaded tokio runtime, with the thread count coming from the main configuration file.

I've also set the sample SetupConfiguration.json to use a default thread count of 4, which I estimate should be enough to prevent bottlenecks for most setups.

I haven't looked deeply into whether this might uncover any thread-safety bugs, but I've ran this with 3000 active client connections without issue. 

I also saw no regression when running against the mongo compatibility test suite. https://github.com/documentdb/documentdb/pull/316


pg_documentdb_gw/src/configuration/mod.rs
```diff
@@ -75,6 +75,9 @@ pub trait SetupConfiguration: DynClone + Send + Sync + Debug {
     /// Returns the time to wait for PostgreSQL to start up before giving up.
     fn postgres_startup_wait_time_seconds(&self) -> u64;
 
+    /// Returns the number of worker threads for the async runtime.
+    fn worker_threads(&self) -> usize;
```
AndrewKhoma: since this setting is specific setting that will be used by Tokio framework, can we name it as `tokio_worker_threads`, to reduce the confusion, please?

pg_documentdb_gw/src/configuration/setup.rs
```diff
@@ -131,4 +134,8 @@ impl SetupConfiguration for DocumentDBSetupConfiguration {
     fn postgres_startup_wait_time_seconds(&self) -> u64 {
         self.postgres_startup_wait_time_seconds.unwrap_or(60)
     }
+
+    fn worker_threads(&self) -> usize {
+        self.worker_threads.unwrap_or(1)
```
visridha: Should we default to core count or something like that?

AndrewKhoma: since it defaults to the number of cpus on the system. source:
https://github.com/tokio-rs/tokio/blob/7c197c7784454b3e78aeddbad102d57f83dd98a4/tokio-macros/src/lib.rs#L64

how about using the https://doc.rust-lang.org/std/thread/fn.available_parallelism.html as the default option?

pg_documentdb_gw/src/main.rs
```diff
@@ -34,23 +33,38 @@ async fn main() {
...
-
+    // Load configuration synchronously
     let setup_configuration = DocumentDBSetupConfiguration::new(&cfg_file)
-        .await
         .expect("Failed to load configuration.");
 
     log::info!(
         "Starting server with configuration: {:?}",
         setup_configuration
     );
```
AndrewKhoma, line 48: this is not your change, but can we have SimpleLogger initialization before this line, please?

pg_documentdb_gw/src/configuration/mod.rs
vtarassov: I actually had it that way originally, but it felt odd to expose that kind of internal detail in the config json and I wanted to keep them consistent. Would you want the config entry to also say "TokioWorkerThreads".

AndrewKhoma: that was my original intention, yes!
from the setup config i want things to be as specific as possible so others can also understand what are we doing
@jcagme what's your thoughts?

jcagme: A little on the fence here and thinking forward. It makes sense to be transparent in terms of what we are using for threading but also it'll require to rename if we decide to move to something like `async_std` in the future. Since today, we are using tokio I think it makes sense to include Tokio in the setting

vtarassov: If we ever do switch runtimes, it seems sub-optimal to require users to make a config change. It seems like the sort of thing that's easy to miss in the patch notes, and might result in a material change in the performance profile. 

I'll draft a version with Tokio verbiage, but I'm also happy to meet in the middle with something like 'RuntimeWorkerThreads' (or 'AsyncRuntimeWorkerThreads')

jcagme: Yeah, exactly my thinking. I'd go with `AsyncRuntimeWorkerThreads` since it's descriptive and yet generic for any runtime changes that might occur

pg_documentdb_gw/src/main.rs
vtarassov: I'm keeping it after the config load for now, with the assumption that it might contain logging configuration in the future.

pg_documentdb_gw/src/configuration/mod.rs
AndrewKhoma: the only realistic one would be for io_uring or rewrite the whole runtime as others do, but yeah, the last one is great

visridha:
@vtarassov  i believe you need to handle DCO for your commits as well

PR #321, opened 2025-09-18
[Draft] Add gateway host extension

visridha:


.gitignore
```diff
@@ -73,4 +73,7 @@ build/
 # deb packages
 *.deb
 
-site/
\ No newline at end of file
+site/
+
+# Rust output
+target/
```
AndrewKhoma, line 79: since we also generate keys, how about adding *.pem here?

pg_documentdb_gw_host/.cargo/config.toml
```diff
@@ -0,0 +1,3 @@
+[target.'cfg(target_os="macos")']
+# Postgres symbols won't be available until runtime
+rustflags = ["-Clink-arg=-Wl,-undefined,dynamic_lookup"]
```
AndrewKhoma, line 3: force-frame-pointers=yes?

PR #335, opened 2025-09-26
Update building.md with simplified workflow

safern:
After trying out the steps in building.md I realized it had some extra steps and some configurations were missing.

This should be simpler and should give a working documentdb instance with Gateway and PostgreSQL extension.

### DIFF
--- a/docs/v1/building.md
+++ b/docs/v1/building.md
@@ -4,14 +4,14 @@ If you want to build and run DocumentDB from source (instead of using Docker), f
 
 ### Prerequisites
 
-Install the required dependencies:
+*Recommended* use the provided devcontainer for VSCode which contains all the dependencies pre-installed.
+
+Or install the required dependencies:
 
 ```bash
 sudo apt update
 sudo apt install build-essential libbson-dev postgresql-server-dev-all pkg-config rustc cargo
 ```
-
-Or you can also use the provided devcontainer configurations for VSCode.
 
 ### Step 1: Build PostgreSQL Extensions
 
@@ -46,12 +46,6 @@ Try basic MongoDB commands to verify everything works.
 ```bash
 psql -p 9712 -d postgres
 ```
-
-### Troubleshooting
-
-- **libbson-static-1.0 not found:**  Install `libbson-dev` and check your `PKG_CONFIG_PATH`.
-- **pg_config not found:**  Install `postgresql-server-dev-all`.
-- **Other errors:**  Double-check all dependencies and your OS version.
 
 ### Need Help?
 


scripts/build_and_install_with_pgrx.sh
```diff
@@ -75,6 +75,15 @@ done
 
 scriptDir="$( cd -P "$( dirname "$source" )" && pwd )"
 
+if [ "$SOURCEDIR" == "" ]; then
+    SOURCEDIR=$scriptDir/../pg_documentdb_gw_host/
```
visridha: Can we keep this script generic - would prefer to have gw specific stuff outside this. This would help tomorrow too if we want a backend Rust extension

scripts/build_and_install_with_pgrx.sh
```diff
@@ -83,7 +92,7 @@ pg_config_path=$pgBinDir/pg_config
 
 # Install cargo-pgrx
 pgrxInstallRequired="false"
-pgrxVersionRequired=$(grep -o 'pgrx = "[^"]*"' $SOURCEDIR/Cargo.toml | cut -d '"' -f 2 | sed 's/=//')
+pgrxVersionRequired=$(grep -o 'pgrx = { version = "[^"]*"' $SOURCEDIR/Cargo.toml | cut -d '"' -f 2 | sed 's/=//')
```
visridha: FYI This would break other callers of this script. Please make the version part optional

scripts/build_and_install_with_pgrx.sh
```diff
@@ -46,21 +49,18 @@ if [ "$help" == "true" ]; then
     echo "  -h                      : Display this help message."
     echo "  -p <package_directory>  : Directory to store the built package (optional)."
     echo "  -r <profile>            : Build profile to use (optional, e.g., release, debug)."
```
AndrewKhoma: not your change, but fyi, the debug profile in Rust is called dev (don't know why, but to be fully correct)

scripts/build_and_install_with_pgrx.sh
AndrewKhoma: ++, i also expect that we might have more than 1 rust extension

scripts/build_and_install_with_pgrx.sh
AndrewKhoma: there are 3 ways of declaring dependency in cargo.toml:
1. [dependencies] section where you say package = "version"
2. [dependencies] section where you say package = { version ="version" }
3. ... you can have a dedicated section for your package like [dependencies.package] and then in the new line you'll have version = "version"
https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#choosing-features

scripts/start_oss_server.sh
```diff
@@ -188,6 +188,7 @@ StartServer $postgresDirectory $coordinatorPort
 
 if [ "$initSetup" == "true" ]; then
   SetupPostgresServerExtensions "$userName" $coordinatorPort $extensionName
+  SetupCustomUser "docdb_user" "Admin100" $coordinatorPort "$userName"
```
AndrewKhoma: see `build_and_start_gateway` it already does it for you

scripts/utils.sh
```diff
@@ -100,6 +100,19 @@ function SetupPostgresServerExtensions()
   psql -p $port -U $user -d postgres -c "SELECT * FROM pg_extension WHERE extname = '$extensionName';"
 }
 
+function SetupCustomUser()
```
AndrewKhoma: maybe we can reuse functionality from `build_and_start_gateway`

AndrewKhoma (commented):
thanks for your changes!

scripts/utils.sh
safern: I think I can move the build_and_start_gateway functionality to utils and cleanup cause we need it here as well.

scripts/build_and_install_with_pgrx.sh
AndrewKhoma: fyi, there's a toml cli tool
https://github.com/gnprice/toml-cli

docs/v1/building.md
```diff
@@ -11,62 +11,42 @@ sudo apt update
...
+#### Using PostgreSQL shell
 
 ```bash
-mongosh --host localhost --port 10260
+psql -p 9712 -d postgres
 ```
 
-Try basic MongoDB commands to verify everything works.
-
 ### Troubleshooting
```
AndrewKhoma: are these troubleshooting still relevant for devcontainer?

visridha (approved):

amitprabhudesai (approved):

PR #336, opened 2025-09-28
Add packaging CI workflow into every PR check and fix rpm errors

shuaitian-git:
- This pull request updates the packaging and CI workflow, run the workflows with certain os/pg for PR checks
- Add new check-extended-rum to package building
- fix a path bug for test-package
- run sql tests with reduced matrix for PR check

### DIFF
--- a/.github/workflows/regress_tests.yml
+++ b/.github/workflows/regress_tests.yml
@@ -41,6 +41,7 @@ jobs:
       pg_version: ${{ matrix.pg_version }}
       arch: ${{ matrix.arch }}
       runner: ${{ matrix.runner }}
+      citus_version: 12
 
   test-full:
     if: github.event_name != 'pull_request'
@@ -64,4 +65,5 @@ jobs:
       pg_version: ${{ matrix.pg_version }}
       arch: ${{ matrix.arch }}
       runner: ${{ matrix.runner }}
+      citus_version: 12
 
--- a/.github/workflows/regress_tests_template.yml
+++ b/.github/workflows/regress_tests_template.yml
@@ -12,6 +12,9 @@ on:
       runner:
         required: true
         type: string
+      citus_version:
+        required: true
+        type: number
 
 jobs:
   test:
@@ -24,7 +27,7 @@ jobs:
       - name: Install package dependencies
         run: |
           export PG_VERSION=${{ inputs.pg_version }}
-          export CITUS_VERSION=12
+          export CITUS_VERSION=${{ inputs.citus_version }}
 
           export LC_ALL=en_US.UTF-8
           export LANGUAGE=en_US


.github/workflows/build_rpm_packages.yml
```diff
@@ -28,16 +25,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [rhel8, rhel9]
+        os: ${{ fromJson(github.event_name == 'pull_request' && '["rhel9"]' || '["rhel8","rhel9"]') }}
```
fatmelon, line 28: because rhel8 and rhel9 use different docker file, not sure we should skip rhel8 ....

shuaitian-git: that should be fine, since those two only differs on postgres repository.

.github/workflows/regress_tests.yml
```diff
@@ -21,10 +21,29 @@ on:
...
+  test-pr:
+    if: github.event_name == 'pull_request' && github.event.pull_request.draft == false
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - pg_version: 15
+            arch: amd64
+            runner: ubuntu-22.04
+          - pg_version: 16
```
xgerman, line 33: are we not building pg16/17 for amd64 - or are we testing arm64 too much?

.github/workflows/regress_tests_template.yml
```diff
@@ -0,0 +1,138 @@
...
+    runs-on: ${{ inputs.runner }}
+    name: Run sql tests on ${{ inputs.runner }} with pg ${{ inputs.pg_version }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Install package dependencies
+        run: |
+          export PG_VERSION=${{ inputs.pg_version }}
+          export CITUS_VERSION=12
```
xgerman: make this an input

.github/workflows/build_deb_packages.yml
```diff
@@ -32,16 +29,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [deb11, deb12, deb13, ubuntu22.04, ubuntu24.04]
+        os: ${{ fromJson(github.event_name == 'pull_request' && '["ubuntu22.04"]' || '["deb11","deb12","deb13","ubuntu22.04","ubuntu24.04"]') }}
```
xgerman, line 32: NIT: make ubuntu verison configutable

.github/workflows/regress_tests.yml
visridha: Primarily this is to reduce hte number of jobs run in PRs while still trying to retain coverage.
I had asked if we could run alternating x64/arm64 runs - with pg18 being tested as arm64 when we add it.

.github/workflows/build_deb_packages.yml
shuaitian-git: do you mean make os configurable when triggered manually(workflow_dispatch)?

the script to build packages already has --os argument.

PR #342, opened 2025-10-06
Add RFC 001: Capped Collections Design

### LINKED ISSUE #341: [Feature Request] Support Capped Collections
**Purpose of the feature.**
Add support for a collection that has circular buffer semantics - when the allocated space is filled, new documents overwrite oldest documents in the collection. 

The user should be able to specify the maximum size of the collection and optionally maximum number of documents for the collection. 

An alternate to this feature is enforcing these constraints on the client.

**Additional context**

nitinahuja89:
This RFC proposes the design of capped collections for DocumentDB.

Closes #341

### DIFF
--- a/rfcs/001-capped-collections.md
+++ b/rfcs/001-capped-collections.md
@@ -1,0 +1,456 @@
+# Design Document: Capped Collections for DocumentDB
+
+## 1. Introduction
+
+This document outlines the design options for implementing MongoDB-compatible capped collections in DocumentDB. Capped collections are fixed-size collections that maintain insertion order and automatically remove old documents when the collection reaches its size limit.
+
+## 2. Capped Collection Behavior
+
+Key behaviors to support:
+- Fixed maximum size with automatic circular reuse of space
+- Optional maximum document count limit
+- Insertion order preservation (natural order)
+- No deletion of documents (only automatic removal when limits are exceeded)
+- No size-increasing updates allowed
+- Support for tailable cursors
+- Support for `convertToCapped` operation
+
+## 3. Out of Scope Features
+
+The following features related to capped collections are out of scope for the current implementation and will be addressed separately:
+
+- **Tailable Cursors**: The ability to create cursors that remain open after returning the last result, allowing for continuous monitoring of a capped collection. This feature will be discussed and implemented separately.
+
+## 4. Topic 1: Enforcing maximum size/ document limits
+
+This topic covers how we will enforce the maximum size and optionally maximum document count limit for capped collections.
+
+### 4.1 Row-Based Trigger
+
+**Description:**
+Trigger fires for each row inserted, checking collection size/document limits and removing oldest documents as needed.
+
+**Overview:**
+This approach implements capped collection functionality by attaching row-level triggers for tables which store capped collection data. Every time a document is inserted, a trigger function executes to check if the collection has exceeded its configured size or document count limits. If limits are exceeded, the trigger removes the oldest documents based on insertion order.
+
+The approach centers around PG's row-level trigger system. After each insert operation the trigger does the following:
+1. Fetch the current collection size and document count from the metadata
+2. Compare against the configured limits
+3. Identify and remove the oldest documents if necessary
+4. Update the collection size and document count in the metadata
+
+We keep the collection size and document count up-to-date in the metadata after each insertion so that it doesn't need to be recomputed each time.    
+This approach allows us to support all the required behaviours for capped collections.
+
+**Pros:**
+- The capped collection is always under the desired limits since the limit is enforced at each document insert
+- Relatively simple implementation using standard PG features
+
+**Cons:**
+- Trigger firing for every row incurs overhead and hence significant performance impact (specially for bulk inserts)
+- Creates dead tuples that would require vacuuming
+
+### 4.2 Statement-Based Trigger
+
+**Description:**
+Triggers fire once after each statement completes, checking collection size/document limits and removing oldest documents in batch if needed.
+
+**Overview:**
+This approach implements capped collection functionality using statement-level triggers that execute once per SQL statement rather than once per inserted row. When a statement completes (potentially inserting multiple documents), a trigger function executes to enforce size constraints.
+
+The implementation uses PG's AFTER STATEMENT trigger which:
+1. Fetch the current collection size and document count from the metadata
+2. Compare against the configured limits
+3. Perform a single cleanup operation that removes enough oldest documents to bring the collection within its limits
+4. Update the collection size and document count in the metadata
+
+This method is more efficient for bulk operations as it performs a single evaluation and cleanup per statement rather than per row. It maintains MongoDB compatibility while offering better performance for common write patterns involving multiple documents.
+
+**Pros:**
+- The capped collection is always under the desired limits since the limit is enforced for each statement
+- Better performance for bulk operations due to efficient batch cleanup of oldest documents
+
+**Cons:**
+- Performance impact for single inserts due to triggers
+- Creates dead tuples that would require vacuuming
+
+### 4.3 Background Worker
+
+**Description:**
+A background process periodically checks capped collections and removes oldest documents to be under the size/document limits.
+
+**Overview:**
+This approach implements capped collection functionality by using a dedicated background worker that runs on a periodic schedule (e.g., every few seconds) to manage collection sizes. Rather than enforcing limits during insert operations, this approach separates the concerns of writing data and maintaining size limits.
+
+The implementation relies on PG's background worker infrastructure, where a scheduled process:
+1. Scans through all capped collections in the system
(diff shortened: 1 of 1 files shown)


rfcs/001-capped-collections.md
```diff
@@ -0,0 +1,456 @@
...
+3. Performs cleanup operations to remove oldest documents as needed
+
+This approach minimizes the performance impact on insert operations since size enforcement happens asynchronously, but may allow collections to temporarily exceed their configured limits between background worker runs.
+
+**Pros:**
+- No impact on insert performance
+- Efficient batch cleanup operations
+
+**Cons:**
+- Potential for temporary size exceedance during high insert throughput and thereby less compatible with MongoDB
```
xgerman, line 97: As long as you are able to order by insert timestamp and then use appropriate limits you will be able to hid the expired entries - so be compatible with Mongo. This adds an additional index...

rfcs/001-capped-collections.md
```diff
@@ -0,0 +1,456 @@
...
+
+The metadata storage approach determines how we track capped collection configuration and statistics.
+
+#### 5.1.1 Extended Collections Table
+
+**Description:**
+This approach extends the existing collections table with additional columns specifically for capped collection properties and statistics.
+
+**Overview:**
+The collections table would be extended to include columns that indicate whether a collection is capped, its maximum size limit, maximum document count, current size, and current document count. This provides a centralized location for all collection metadata, keeping capped collection configuration alongside standard collection information.
```
xgerman, line 265: I don't think we need current size and current document count for all approahes, e.g. a background worker could first determine the size and then decide to act...

rfcs/001-capped-collections.md
```diff
@@ -0,0 +1,456 @@
...
+**Cons:**
+- Requires joins for some operations that need both regular and capped collection metadata
+- More complex schema management with an additional table
+
+#### Recommendation
+I would recommend having a separate capped collection metadata table to avoid concurrency bottlenecks on the primary collection metadata table when capped collection statistics are updated. 
+
+### 5.2 Insertion Order Tracking
+
+Tracking insertion order is essential for capped collections to ensure natural order queries and document cleanup.
```
xgerman, line 299: do we need order or is timestamp sufficicient aka let's assume several documents are inserted at the same time and we remoive the wrong document because we didn't record order - how big an issue is that?

rfcs/001-capped-collections.md
```diff
@@ -0,0 +1,456 @@
...
+
+**Pros:**
+- Simple, built-in PG feature with minimal implementation effort
+- Easy to query and index
+- Automatically maintained by the database system
+- Handles concurrent inserts properly
+
+**Cons:**
+- The sequence generator may become a concurrency bottleneck for high-volume inserts
+- Less control over value assignment and sequence management
```
xgerman, line 317: also whatver sequence we use it might eventually wrap around and start with 0 again...

rfcs/001-capped-collections.md
nitinahuja89: I don't completely understand the question. We need to track insertion order for both queries looking for natural ordering and cleanup. With timestamp we will need to guarantee there are no two records with the same timestamp. Also, what is the motivation behind using timestamp instead of a big serial?

rfcs/001-capped-collections.md
nitinahuja89: Yes we will need to reset it to 0 when it has hit the limit

rfcs/001-capped-collections.md
nitinahuja89: Yes that's true. It's not as important for the background worker approach.

rfcs/001-capped-collections.md
nitinahuja89: There will be a delay in hiding the appropriate entries since this will be managed by an asynchronous background worker.

xgerman: Like you do with insert, update, delete you can also modify retrival to only return the relevant documents. It's probably something you are duiscounting due to the hit on the read path - so just would like to see it for completness aka in the case I was thinking you would add to the read path something like `select * from X order by Y desc LIMIT <collection size>` and then operate on that... you might be able to make that a view?

rfcs/001-capped-collections.md
xgerman: Mostly  curiosity aka if we are ok with evicting occasionally the "wrong" document that opens up us not having to synchronize on the sequence which might be useful in a sharded envrionment...

rfcs/001-capped-collections.md
```diff
@@ -0,0 +1,456 @@
...
+3. **Hybrid Approach**:
+   - New partitions are created on-demand during the insert when an insertion would exceed current partition boundaries
+   - A separate scheduled process (pg_cron or background worker) runs independently to enforce size limits by dropping old partitions
+   - This hybrid solution balances immediate insertion performance with efficient batch cleanup by separating the two concerns
+
+If we go with table partitioning, I would recommend the hybrid approach since it will guarantee that the critical task of creating new partitions when required happens during insertion (so that insertions always succeed) and we can hand-off the less critical task of dropping old partitions to a separate scheduled process. 
+
+Key considerations for partition management include:
+- **Partition Size**: Smaller partitions allow more precise size control but create management overhead
+- **Boundary Selection**: Parition boundaries need to be decided based on partition size and optionally document count
```
jcagme, line 139:
```suggestion
- **Boundary Selection**: Partition boundaries need to be decided based on partition size and optionally document count
```

rfcs/001-capped-collections.md
```diff
@@ -0,0 +1,456 @@
...
+- Minimal performance overhead
+- Provides specific, MongoDB-compatible error messages
+
+**Cons:**
+- Needs additional context information about the collection (check if capped collection) to enforce the constraint
+- Must be carefully maintained ensuring that all delete path have this check
+
+#### Recommendation
+
+I recommend that we integrate the contraint check in the core delete functions since it allows providing specific MongoDB compatible error messages with minimal performance overhead.
```
jcagme, line 418:
```suggestion
I recommend that we integrate the constraint check in the core delete functions since it allows providing specific MongoDB compatible error messages with minimal performance overhead. 
```

jcagme (commented):
Some typos

rfcs/001-capped-collections.md
nitinahuja89: Even if we think of read path skipping expired entries, we still need some mechanism which marks the entries as expired in the first place. What is that mechanism which marks the appropriate documents as expired?

rfcs/001-capped-collections.md
nitinahuja89: I think it's important to have a deterministic behavior where the oldest entries get removed first to cater to a wide variety of applications using this.

rfcs/001-capped-collections.md
```diff
@@ -0,0 +1,456 @@
...
+
+| Criteria | Row-Based Trigger | Statement-Based Trigger | Background Worker | Partitioning | Transaction Hook | Custom Table Access Method |
+|----------|-------------------|--------------------------|-------------------|--------------|------------------|----------------|
+| MongoDB Compatibility | High | High | Medium | Low | High | High |
+| Performance | Low | Medium | High | Very High | High | Very High |
+| Vacuuming for deleted documents | Yes | Yes | Yes | No | Yes | No |
+| Implementation Complexity | Medium | Medium | Medium | High | High | Very High |
+| Maintenance | Simple | Simple | Medium | Complex | Complex | Very Complex |
+
+The ideal approach would be to implement a specialized storage engine via a custom table access method. However, that will have a much higher implementation and maintenance complexity than any other approach. We could evaluate this approach further to assess the cost-benefit of this approach. The next best option is to use a custom transaction hook approach since it provides good performance in addition to MongoDB compatibility at a reasonable implementation and maintenance cost.
```
visridha, line 249: Would be good to add a summary of what exactly the proposal is (Is the idea to choose a custom transaction hook?)

rfcs/001-capped-collections.md
```diff
@@ -0,0 +1,456 @@
...
+- Separates concerns, keeping regular collections metadata unaffected
+- Can include more specialized fields for capped collections
+- Better for horizontal scaling and concurrency
+
+**Cons:**
+- Requires joins for some operations that need both regular and capped collection metadata
+- More complex schema management with an additional table
+
+#### Recommendation
+I would recommend having a separate capped collection metadata table to avoid concurrency bottlenecks on the primary collection metadata table when capped collection statistics are updated.
```
visridha, line 295: agreed

rfcs/001-capped-collections.md
nitinahuja89: Do you mean summary of the transaction hook approach?

PR #361, opened 2025-10-24
[DoNotMerge] testing

shuaitian-git:


.github/workflows/gateway_tests.yml
```diff
@@ -11,13 +11,7 @@ on:
       - "main"
     paths:
       - "pg_documentdb_gw/**"
-  pull_request:
```
AndrewKhoma, line 14: why are we removing the check for PRs here?

shuaitian-git: testing only

Issue #60, opened 2025-02-12
[Feature Request] `killOp` command


AlekSi:
**Purpose of the feature.**

#52 fixed `$currentOp` aggregation pipeline stage. It would be nice if DocumentDB also provided a safe way to stop the operation – `killOp` command.

**Describe the solution you'd like**

DocumentDB implements `killOp` command.

**Describe alternatives you've considered**

Implement it in FerretDB (somehow).

**Additional context**

N/A.

diipak-bisht:
@AlekSi, I am working on this. it should land up hopefully in 0.108

diipak-bisht:
The target release date for kill op is shifted to 0.109

PR #365, opened 2025-10-31
DocumentDB sync 20251031

### LINKED ISSUE #359: $addToSet operator in update workflow inserting more data than expected when used with LibreChat AI framework.
**Describe the bug**

Very detailed description and repro steps found here: https://github.com/danny-avila/LibreChat/discussions/8596

Appending the last part of the discussion for reference:

>figured out the issue (with a lot of help from Claude Sonnet 4.5). The issue does indeed stem from Cosmos DB > for MongoDB. We have to adjust the `addAgentResourceFile` function in the `Agent.js` file to NOT use `$addToSet`.
> 
> Original function:
> 
> ```
> const addAgentResourceFile = async ({ req, agent_id, tool_resource, file_id }) => {
>   const searchParameter = { id: agent_id };
>   let agent = await getAgent(searchParameter);
>   if (!agent) {
>     throw new Error('Agent not found for adding resource file');
>   }
>   const fileIdsPath = `tool_resources.${tool_resource}.file_ids`;
>   await Agent.updateOne(
>     {
>       id: agent_id,
>       [`${fileIdsPath}`]: { $exists: false },
>     },
>     {
>       $set: {
>         [`${fileIdsPath}`]: [],
>       },
>     },
>   );
> 
>   const updateData = {
>     $addToSet: {
>       tools: tool_resource,
>       [fileIdsPath]: file_id,
>     },
>   };
> 
>   const updatedAgent = await updateAgent(searchParameter, updateData, {
>     updatingUserId: req?.user?.id,
>   });
>   if (updatedAgent) {
>     return updatedAgent;
>   } else {
>     throw new Error('Agent not found for adding resource file');
>   }
> };
> ```
> 
> Updated function:
> 
> ```
> const addAgentResourceFile = async ({ req, agent_id, tool_resource, file_id }) => {
>   const searchParameter = { id: agent_id };
>   
>   // Get current agent
>   let agent = await getAgent(searchParameter);
>   if (!agent) {
>     throw new Error('Agent not found for adding resource file');
>   }
> 
>   // Build the updated arrays in JavaScript, not in Cosmos DB
>   const currentTools = agent.tools || [];
>   const updatedTools = currentTools.includes(tool_resource) 
>     ? currentTools 
>     : [...currentTools, tool_resource];
> 
>   const currentFileIds = agent.tool_resources?.[tool_resource]?.file_ids || [];
>   const updatedFileIds = currentFileIds.includes(file_id)
>     ? currentFileIds
>     : [...currentFileIds, file_id];
> 
>   // Use $set instead of $addToSet to avoid Cosmos DB bugs
>   const updateData = {
>     $set: {
>       tools: updatedTools,
>       [`tool_resources.${tool_resource}.file_ids`]: updatedFileIds,
>     },
>   };
> 
>   const updatedAgent = await updateAgent(searchParameter, updateData, {
>     updatingUserId: req?.user?.id,
>     skipVersioning: true,
>   });
> 
>   if (updatedAgent) {
>     return updatedAgent;
>   } else {
>     throw new Error('Agent not found for adding resource file');
>   }
> };
> ```
> 
> The interesting thing here is that Cosmos DB for MongoDB [appears to support the $addToSet operator](https://learn.microsoft.com/en-us/azure/cosmos-db/mongodb/vcore/operators/array-update/$addtoset). However, it doesn't seem to be working. I've created a ticket with Microsoft support to report the bug, but, even if they confirm it, I suspect it will be a while before this is fixed.
> 
> So...the big question for @danny-avila is, would you support refactoring LibreChat to not use `$addToSet`? There aren't too many references to this, so I believe this could be done fairly easily, and I'd be happy to submit that PR. Obviously, we can always go to running actual MongoDB on a VM, but we would prefer to not go that route if possible. Atlas is likely not an option for us, as we do want to keep everything within infrastructure we control.
> 
> Unless someone else can look at this and determine that it isn't actually the `$addToSet` operator but something else which is causing the issue?

shuaitian-git:
closes #359


shuaitian-git:
> > closes #359
> 
> this sync would fix this issue as well. #60

Deepak mentioned there is another integration work with GW.

sivethe (approved):

Issue #371, opened 2025-11-06
[Feature Request] Add https support to mongodb-local


xgerman:
**Purpose of the feature.**
Dealing with (self-signed) certificates can be confusing and difficult, espeically in CI/CD environments

**Describe the solution you'd like**
Support a non TLS mode

**Describe alternatives you've considered**
N/A

**Additional context**
This needs to be optional so docuemntb-local remains secure by default.

AndrewKhoma:
@vtarassov is working on PR for this one

PR #373, opened 2025-11-11
[Gateway] Adding support for non-TLS connections

vtarassov:
### Does this PR have any customer impact?
No.

### Type (Feature, Refactoring, Bugfix, DevOps, Testing, Perf, etc)
Feature

### Does it involve schema level changes? (Table, Column, Index, UDF, etc level changes)
No

### Are you introducing any new config? If yes, do you have tests with and without them being set?
Yes. 

### ChangeLog (Refer [Template](../oss/CHANGELOG.md))
Support non-TLS connections

### Description
Issue: https://github.com/documentdb/documentdb/issues/371

Added a new 'EnforceTls' parameter that, if false, will allow clients to connect without TLS. Default is 'true'.

When TLS enforcement is disabled, the gateway will peek the first three bytes of the stream to see if they match the TLS handshake and establish a corresponding connection. The client has a limited amount of time to send the first three bytes (5s currently) before the connection is closed by the gateway. 

This is primarily intended for development purposes and should not be used in production. As such, there is no option to disable or disallow TLS.

### DIFF
--- a/pg_documentdb_gw/src/configuration/setup.rs
+++ b/pg_documentdb_gw/src/configuration/setup.rs
@@ -26,6 +26,7 @@ pub struct DocumentDBSetupConfiguration {
     // Gateway listener configuration
     pub use_local_host: Option<bool>,
     pub gateway_listen_port: Option<u16>,
+    pub enforce_tls: Option<bool>,
 
     // Postgres configuration
     pub postgres_system_user: Option<String>,
@@ -139,5 +140,9 @@ impl SetupConfiguration for DocumentDBSetupConfiguration {
 
     fn postgres_idle_connection_timeout_minutes(&self) -> u64 {
         self.postgres_idle_connection_timeout_minutes.unwrap_or(5)
+    }
+
+    fn enforce_tls(&self) -> bool {
+        self.enforce_tls.unwrap_or(true)
     }
 }
--- a/pg_documentdb_gw/src/lib.rs
+++ b/pg_documentdb_gw/src/lib.rs
@@ -30,7 +30,7 @@ use socket2::TcpKeepalive;
 use std::net::IpAddr;
 use std::{pin::Pin, sync::Arc, time::Duration};
 use tokio::{
-    io::BufStream,
+    io::{AsyncRead, AsyncWrite, BufStream},
     net::{TcpListener, TcpStream},
 };
 use tokio_openssl::SslStream;
@@ -58,7 +58,8 @@ const TCP_KEEPALIVE_INTERVAL_SECS: u64 = 60;
 const STREAM_READ_BUFFER_SIZE: usize = 8 * 1024;
 const STREAM_WRITE_BUFFER_SIZE: usize = 8 * 1024;
 
-type GwStream = BufStream<SslStream<TcpStream>>;
+// TLS detection timeout
+const TLS_PEEK_TIMEOUT_SECS: u64 = 5;
 
 /// Runs the DocumentDB gateway server, accepting and handling incoming connections.
 ///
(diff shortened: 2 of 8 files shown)


jcagme (approved):
LGTM. Thanks for the change

pg_documentdb_gw/src/lib.rs
```diff
@@ -127,11 +128,88 @@ where
...
+                );
+                return Err(DocumentDBError::internal_error(format!(
+                    "Error reading from stream {e:?}"
+                )));
+            }
+            Err(_) => {
+                log::warn!(
+                    activity_id = connection_id.to_string().as_str();
+                    "TLS detection peek operation timed out after {} seconds.",
+                    TLS_PEEK_TIMEOUT_SECS
```
AndrewKhoma, line 197: this const should go inside of the format msg (run `cargo fmt --all --check` and `cargo clippy --all-targets --all-features -- -D warnings` on the pg_documentdb_gw folder)

vtarassov: I wasn't able to get the gateway tests running locally via act on my arm machine, but I ran these commands locally and they seemed to pass. Let's see what the workflows think.

pg_documentdb_gw/src/lib.rs
```diff
@@ -127,11 +128,87 @@ where
...
+                    "TLS detection peek operation timed out after {TLS_PEEK_TIMEOUT_SECS} seconds."
+                );
+                return Err(DocumentDBError::internal_error(
+                    "Timeout reading from stream".to_string(),
+                ));
+            }
+        }
+
+        // Successive peeks to a non-empty buffer will return immediately, so we wait before retry.
+        tokio::time::sleep(Duration::from_millis(100)).await;
```
visridha, line 206: Is 100ms a safe time for this?

pg_documentdb_gw/src/lib.rs
```diff
@@ -234,15 +345,15 @@ where
                 let request_activity_id =
                     connection_context.generate_request_activity_id(header.request_id);
 
-                if let Err(e) = handle_message::<T>(
+                if let Err(e) = handle_message::<T, _>(
```
visridha: Generally am a fan of being more specific where possible - any reason to prefer `_` over just using `S` here?

pg_documentdb_gw/src/configuration/setup.rs
```diff
@@ -28,6 +28,7 @@ pub struct DocumentDBSetupConfiguration {
     // Gateway listener configuration
     pub use_local_host: Option<bool>,
     pub gateway_listen_port: Option<u16>,
+    pub enforce_tls: Option<bool>,
```
visridha, line 29: Should we seed the json file?

pg_documentdb_gw/src/lib.rs
vtarassov: No, I can make the change.

pg_documentdb_gw/src/lib.rs
vtarassov: What are the risks you're concerned about?

This should only be hit in the scenario where for whatever reason we didn't get all three TLS handshake bytes on the first tcp frame. This might occur naturally in some niche networking edge cases where the frame gets split in transit. In that case, this sleep would mean a delay in connecting and we wouldn't want it too high.

Otherwise, if the client is doing it intentionally and/or maliciously, peek should be fairly well optimized on modern OS's and the perf overhead should be limited. After the first invocation it should be reading from a memory-mapped buffer and shouldn't need a full context switch. 

That being said, I haven't load-tested this scenario so I can set it to 500ms or 1s if you're more comfortable with that.

pg_documentdb_gw/src/configuration/setup.rs
vtarassov: I could go either way. I opted not to since this isn't a feature we'd encourage in production but no strong opinions.

visridha: Ah gotcha - then yeah leaving it off makes sense for non-default setups.

pg_documentdb_gw/src/lib.rs
visridha: yeah i subsequently got to the point where this wasn't default - so i'm good with leaving it as-is for now.

visridha (approved):

PR #384, opened 2025-12-03
Add Rfc0008 - Introduces a version naming scheme

xgerman:


rfcs/0008-versioning_scheme.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Approach
+
+Currently we are running `0.106-1` with `0` being the major version, `106` being the minor
+version, an `-1` the patch version. There is some loose understanding that once we reach a 
+certain maturity or if there is a breaking change we will go to major version `1`.
+
+Going forward we will codify things as follows:
+
+* Every year in November we will release a long term support version which will be supported
```
nitinahuja89, line 30: How long after a major version is released will we release a long term support version for it? For example, we release a new major version in June and by November we release a long term support version for that major version.

rfcs/0008-versioning_scheme.md
```diff
@@ -0,0 +1,145 @@
...
+  for (at least) one year with  patch versions aka bug fixes and security fixes - which get backported. Each new release will have the same major version as used during that year and the minor version will be `100`, e.g. `1.100-0`. Each subsequent patch will increment the patch version, e.g. `1.100-1`.
```
nitinahuja89, line 31: So this means that after a major version has a LTS version, we fix the minor version at 100 and only increment the patch versions, is that what you are trying to say? There would be no more minor versions released for that major version once the LTS version for that major is released - is that correct?

rfcs/0008-versioning_scheme.md
```diff
@@ -0,0 +1,145 @@
...
+* During the year we will work on the new major version and have minor version releases
```
nitinahuja89, line 32: Should we classify what some examples on which changes would be released as a minor version vs what changes would be released as patch version?

rfcs/0008-versioning_scheme.md
xgerman: That's correct. We are looking for a predictable release schedule so we aim to release a new LTS version *every* November. Working backward June is not a bad guess for a code freeze - but the mechanics of releasing are not subject of theis RFC (we have debate if it should be though)

rfcs/0008-versioning_scheme.md
xgerman: This is correct. LTS versions don't get new features, only bug/security fixes so only patch version should increase. I can't imagine a scenario where a bug/security fix would increase the minor version but if that would occur we woudl take it.

normtown:
What needs to happen to close on this? When can we expect a new revision? We presumably want to release a 1.0 in 2026.

rfcs/0008-versioning_scheme.md
alaye-ms: The idea here is after the LTS version of (for example) 3 is released as 3.100-0, then the very next release will be 4.0 in November, incrementing until next November of the next year. I don't think we need a code freeze, since the LTS version is frozen for the whole year anyway.

rfcs/0008-versioning_scheme.md
alaye-ms: I added in a modified version of this PR, please feel free to look https://github.com/documentdb/documentdb/pull/597

alaye-ms:
I took over this PR and added some more details here: https://github.com/documentdb/documentdb/pull/597

PR #386, opened 2025-12-04
Add RFC-0007: Guidance for Onboarding Statistics

WentingWu666666:
## Summary

This PR establishes standardized conventions for contributors to add new statistics to DocumentDB.

## Motivation

DocumentDB currently lacks consistent guidance for exposing runtime, performance, and usage statistics. This leads to fragmented implementations, inconsistent APIs, and friction for both contributors and reviewers.

## What's Included

- **Naming conventions**: Standardized patterns for views (`documentdb_stat_<scope>`), helper functions, and reset functions
- **Column conventions**: Required unit suffixes (`_count`, `_seconds`, `_bytes`, etc.) for value columns, suffix-free dimension columns
- **Permission model**: Public SELECT for views, restricted EXECUTE for reset functions
- **Configuration flags**: `documentdb_track_<scope>` to enable/disable collection with minimal overhead
- **Documentation requirements**: Mandatory updates to stats.md for each new statistic

## Non-Goals

- Does not implement specific statistics
- Does not modify PostgreSQL's existing statistics system
- Does not address query planner optimizations

### DIFF
--- a/rfcs/0007-stats.md
+++ b/rfcs/0007-stats.md
@@ -1,0 +1,182 @@
+---
+rfc: 0007
+title: "RFC Title"
+status: Draft
+owner: "@github-username"
+issue: "https://github.com/documentdb/documentdb/issues/XXX"
+discussion: "https://github.com/documentdb/documentdb/discussions/XXX"
+version-target: 1.0
+implementations:
+  - "https://github.com/documentdb/documentdb/pull/XXX"
+---
+
+# RFC-XXXX: [Title]
+
+*Delete these italicized instructions before submitting your RFC. For details of the RFC process, see https://github.com/documentdb/documentdb/blob/main/rfcs/0003-rfc-process.md*
+
+*RFC statuses follow this flow:*
+
+```
+[Draft] → [Proposed] → [Accepted/Rejected] → [Implementing] → [Complete]
+                                           ↘ [Archived]
+```
+
+## Problem
+
+*This section is REQUIRED before moving from Draft to Proposed status.*
+
+**Section purpose:** Clearly articulate the problem or opportunity this RFC addresses. 
+
+**Complete this section when:** You're ready to start a discussion and get initial feedback on whether this problem is worth solving.
+
+**Guidance:**
+- What problem are you trying to solve?
+- Who is impacted by this problem?
+- What are the consequences of not solving it?
+- What are the current workarounds, if any?
+- What are the success criteria?
+- What are the non-goals?
+- What behavior are you intending to emulate? What are the quircks there?
+- Is there an API spec being targeted?
+
+**Example prompts to guide your thinking:**
+- "Users currently struggle with..."
+- "The system lacks..."
+- "This creates friction when..."
+- "Without this, contributors must..."
+
+---
+
+## Approach
+
+*This section is REQUIRED before moving from Proposed to Accepted status.*
+
+**Purpose:** Describe your proposed solution at a high level.
+
+**Complete this section when:** You've received feedback that the problem is worth solving and you're ready to propose a specific approach.
+
+**Progressive disclosure:** Start lean! You don't need all the details yet. Focus on the core idea and high-level approach. Detailed design comes later.
+
+**Guidance:**
+- What is your proposed solution?
+- Why is this approach better than alternatives?
+- What are the key benefits and tradeoffs?
+- How does this fit with existing DocumentDB architecture?
+
+**Example prompts for solution thinking:**
+- "The proposed solution is to..."
+- "This approach is preferable because..."
+- "Key tradeoffs include..."
+- "This aligns with existing patterns by..."
+
+---
+
+## Detailed Design
+
+*This section MAY BE REQUIRED before moving from Proposed to Accepted status. This section MUST be completed and approved to move to Implementing status.*
+
+**Purpose:** Provide comprehensive technical details needed for implementation.
+
+**Complete this section when:** Your solution approach has been validated and you're ready to commit to specific implementation details.
+
+**Guidance:** This is where you get specific. Include enough detail that someone could implement this RFC without having to make major design decisions.
+
+### Technical Details
+
+*Describe the technical implementation specifics*
+- Data structures
+- Algorithms
+- Architecture patterns
+- Performance considerations
+
+### API Changes
+
+*Document any public API additions or modifications*
+- New functions, including UDFs
+- Modified signatures
+- Breaking changes
+- Deprecation plans
+
+### Database Schema Changes
+
+*If applicable, describe schema modifications*
+- New tables/collections
+- Schema migrations
+- Index changes
+- Data migration strategies
+
+### Configuration Changes
+
+*Document new or modified configuration options*
+- New settings
+- Modified defaults
+- Environment variables
+- Configuration validation
+
+### Testing Strategy
+
+*Describe how this will be tested*
+- Unit test approach
+- Integration test requirements
+- Compatibility test requirements
+- Performance test plans
+- Migration test strategy
+
+### Migration Path
+
+*How do existing users/deployments upgrade?*
+- Backwards/forwards compatibility
+- Migration steps
+- Rollback strategy
+- Deprecation timeline
+
+### Documentation Updates
+
+*What documentation needs to change?*
+- User-facing docs
+- Developer guides
+- API references
+- Examples/tutorials
+
+---
+
+## Implementation Tracking
+
+*This section SHALL be populated during the Implementation phase.*
+
+**Purpose:** Track the implementation progress of this RFC.
+
+**Complete this section when:** Your RFC has been accepted and implementation work begins.
+
+**Guidance:**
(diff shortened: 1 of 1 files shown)


jayanta-mondal:
What is an example of a statistic here? Like the metrics we emit?

Can you add some read code examples to the doc. Otherwise it's hard to know if this is a real problem that needs to be addressed.

jayanta-mondal:
Can you please elaborate about what you mean by permission model? What is the view here? An abstract view of the statistics - that is not yet implemented?

WentingWu666666:
> What is an example of a statistic here? Like the metrics we emit?
> 
> Can you add some read code examples to the doc. Otherwise it's hard to know if this is a real problem that needs to be addressed.

 Thanks @jayanta-mondal! Updated the RFC with:
 
  - Expanded **Problem** section with background & motivation (what exists today, what's missing, why conventions matter now) — kept the template heading but added the context you asked for so it's self-contained
  - **Contributor Checklist** with code examples for each step (GUC, helper, view, reset, grants, tests) using a hypothetical io scope
 
This RFC proposes conventions, not a specific statistic — the io examples are illustrative.

WentingWu666666:
> Can you please elaborate about what you mean by permission model? What is the view here? An abstract view of the statistics - that is not yet implemented?

"view" means a literal CREATE VIEW in `__API_CATALOG_SCHEMA__` that users `SELECT` from. 

Permission model: `SELECT TO PUBLIC` on views, reset functions restricted to `__API_ADMIN_ROLE__`, helpers private in `__API_SCHEMA_INTERNAL__`. 

Both explained upfront in the **Approach** section, with a macro glossary before first use in API Changes

rfcs/0007-stats.md
```diff
@@ -0,0 +1,432 @@
...
+
+- This RFC does **not** design or implement any specific statistics.
+- This RFC does **not** replace or modify existing PostgreSQL statistics.
+- This RFC does **not** address using statistics for query planning or optimization. The focus is solely on exposability and monitoring.
+
+---
+
+## Approach
+
+The approach is inspired by established conventions in PostgreSQL (e.g., `pg_stat_*` views) and widely-used extensions (e.g., `pg_stat_statements`).
```
jayanta-mondal, line 100: CREATE VIEW pg_stat_statements AS
  SELECT * FROM pg_stat_statements(true);

So, in PG the view is pass through - which makes sense - because view is tied to a postgres instance - it won't know distribution, workers etc. 

So - it's not the view - it's the function and the return type we need to standardize.  the view is just a skin on top.

jayanta-mondal: And the function should return a bson[] instead of a fix schema. 

bson[] array will be immune to migration. 

A schematized view or function will be harder to updrage/migrate. 

We want the upgrade path to be as streamlined as possible.

rfcs/0007-stats.md
```diff
@@ -0,0 +1,432 @@
...
+  - `_bytes`
+  - `_percent`
+
+- Columns representing **dimensions** (e.g., name, database, collection, user) should **not** use a suffix.
+
+- Metadata columns that carry a timestamp are exempt from the unit-suffix rule. Specifically, `stats_reset` is required if the view contains cumulative counters and indicates the timestamp of the last reset. Other timestamp-typed metadata columns (e.g., `last_updated`, `first_seen`) are likewise exempt.
+
+**Example** (for illustration purposes only, not an actual view):
+```sql
+CREATE VIEW documentdb_stat_queries AS
```
jayanta-mondal, line 167: As mentioned before - the view should be a wrapper over a function that returns a bson[]

rfcs/0007-stats.md
```diff
@@ -0,0 +1,432 @@
...
+pg_documentdb/sql/udfs/stats/stats--latest.sql
+pg_documentdb/sql/udfs/stats/stats--<from>-<to>.sql
+```
+
+#### Helper Functions
+If a view is backed by helper functions, those functions must follow this naming pattern:
+
+- Single helper per scope:
+  ```
+  __API_SCHEMA_INTERNAL__.documentdb_stat_get_<scope>
```
jayanta-mondal, line 202: They should be the core workforce not helper 

if you we follow the PG pattern

rfcs/0007-stats.md
```diff
@@ -0,0 +1,432 @@
...
+issue: "https://github.com/documentdb/documentdb/issues/TBD"
+---
+
+# RFC-0007: Guidance for Onboarding Statistics
+
+## Problem
+
+### What exists today
+
+DocumentDB exposes diagnostic commands such as `coll_stats`, `db_stats`, `index_stats`, and `current_op`. These exist to provide **MongoDB API compatibility** — they return BSON documents and mirror the behavior of their MongoDB counterparts. They are not designed as a reusable pattern for exposing new PostgreSQL-native statistics.
```
jayanta-mondal, line 15: so coll_stats is client facing but pg_stat_* views are service-provider facing 

Will the statistics be client facing?

If it's client facing - what guidance/mechanism we will have for what statistic can be exposed to client vs not.

rfcs/0007-stats.md
```diff
@@ -0,0 +1,432 @@
...
+
+Add the view, optional helper(s), and optional reset function under `pg_documentdb/sql/udfs/stats/`:
+
+**Helper function** (in `stats--latest.sql`):
+
+```sql
+CREATE OR REPLACE FUNCTION __API_SCHEMA_INTERNAL__.documentdb_stat_get_io()
+RETURNS TABLE (
+    database       text,
+    read_count     bigint,
```
jayanta-mondal, line 388: What guidance/gate will have for performance of these statics collection wrt. 
1. hot path imapact
2. throughput etc. 

For example - how should one think about if a new statistic emission won't regress performance.

rfcs/0007-stats.md
WentingWu666666: I've added a scope clarification in Non-goals — this RFC covers extension-defined statistics consumed via SQL (psql, Grafana, monitoring dashboards), not stats exposed through the MongoDB wire protocol.
 
 Given that scope:
  - View backing: views don't always need a function — if the stat comes from a table or a simple SQL query, the view is the interface directly. The RFC covers both patterns: function-backed (shared memory) and SQL-backed (tables).
  - Typed schema vs bson[]: for SQL consumers, typed columns give you native WHERE/JOIN, information_schema discoverability, and compatibility with standard PG monitoring tools. Every pg_stat_* uses typed schemas for this reason. Adding columns is non-breaking with CREATE OR REPLACE VIEW and our existing upgrade scripts.
 
 If there's a future need for stats exposed through the wire protocol, that would warrant a separate RFC with bson[] conventions. This RFC is scoped to SQL-only consumption.

rfcs/0007-stats.md
WentingWu666666: Same reasoning as above — since these stats are consumed via SQL only, typed columns are a better fit than bson[]. The view can be backed by a function (typed return), a table, or a plain SQL query depending on the statistic's needs.

rfcs/0007-stats.md
WentingWu666666: The RFC calls them "helpers" not because they're unimportant, but because they're not the public API. The view is what consumers query — the function behind it is an implementation detail that lives in `__API_SCHEMA_INTERNAL__` and is not granted to PUBLIC.
 
Why this matters: if the function signature is private, we can change it freely across versions (add/remove parameters, rename it) without breaking any consumer. The view absorbs those changes. If we expose the function directly as the API, its signature becomes frozen.
 
 This matches how PostgreSQL does it internally — pg_stat_get_activity() is the internal function, pg_stat_activity is the public view. Users query the view; the function is an implementation detail.

rfcs/0007-stats.md
WentingWu666666: No — added a Non-goals clarification: this RFC covers only extension-defined statistics consumed via SQL (psql, Grafana, monitoring dashboards). Not exposed through the MongoDB wire protocol. coll_stats/db_stats are client-facing and follow MongoDB API conventions; the stats in this RFC are a separate category.

rfcs/0007-stats.md
WentingWu666666: Good call — I've added a "Performance Considerations" section to the RFC covering:
 
  1. Lock-free collection (atomic ops, per-backend buffers) for counters in the query execution path
  2. GUC must fully disable collection (zero overhead when off), not just hide output
  3. Memory budget declared in PR for shared-memory stats
  4. Before/after benchmark required showing negligible throughput regression
 
 Let me know if that covers what you had in mind.

WentingWu666666:
Hi @jayanta-mondal @safern, I’ve addressed Jayanta’s feedback and clarified the prototype/broader review questions in Teams. My understanding is that this RFC defines initial conventions only and does not ship a public stats interface yet. The proposed interface follows existing PostgreSQL stats patterns, and we can still refine the convention before the first real statistic ships if implementation exposes issues.

Could you please let me know if there are any remaining blockers? If this looks reasonable, could you approve so we can move forward with merging?

jayanta-mondal (approved):

PR #408, opened 2026-01-06
Add Unix domain socket support for gateway

### LINKED ISSUE #400: [Feature Request] Add Unix Domain Socket Support
**Purpose of the feature.**
Enable the gateway to accept connections via Unix domain sockets in addition to TCP/TLS. Currently, all connections must go through TCP networking, even when the client application runs on the same machine as the gateway. This introduces unnecessary overhead for local communication scenarios.

Unix domain sockets provide a more efficient IPC (Inter-Process Communication) mechanism for same-machine connections, offering better performance, enhanced security, and simplified configuration for deployments where the client and gateway run on the same host.

Add support for Unix domain socket listeners alongside the existing TCP listener.  When enabled, clients running on the same host should be able to connect using a local socket file (for example, /tmp/osddb.sock) instead of TCP. This feature should be optional and fully backward-compatible with existing TCP-based configurations.

**Configuration:**
Unix domain socket support should be controlled through configuration flags.

```json
{
  "UnixSocketEnabled": true,
  "UnixSocketPath": "/tmp/osddb.sock"
}
```

**UnixSocketEnabled (boolean):**

When set to true, the gateway creates and listens on a Unix domain socket.

When set to false or omitted, Unix socket support is disabled and the gateway behaves exactly as it does today (TCP-only).

**UnixSocketPath (string, optional):**

Specifies the filesystem path for the Unix socket.

If not provided, a default path such as /tmp/osddb.sock is used.

This field is only relevant when UnixSocketEnabled is set to true.

**Implementation approach:**
- Add a configurable `unix_socket_enabled` flag and `unix_socket_path` parameter to the gateway configuration.
- Create Unix socket listener that accepts connections concurrently with TCP
- Reuse existing MongoDB protocol handling code for both connection types
- Set appropriate file permissions (0o660) on socket file for security
- No TLS required for Unix sockets (local-only, file permissions provide security)

**Key benefits:**

1. **Security:** Enhanced security for local connections
   - No network exposure
   - File system permissions control access
   - No certificates needed for local connections

2. **Operational simplicity:**
   - No port conflicts for local connections
   - Standard Unix IPC pattern

Alternative inter-process communication mechanisms were considered for local connections. One option is shared memory, where multiple processes communicate by reading from and writing to a common memory region. Another option is named pipes, which allow processes to exchange data through special filesystem-based endpoints.

While both approaches can support local communication, they introduce additional complexity and are less commonly used for database client connectivity. In contrast, Unix domain sockets are a well-established and widely adopted solution for local IPC in database systems. Major databases such as PostgreSQL, MySQL, and MongoDB rely on Unix domain sockets for local connections.

**Additional context**

This feature is intended to be optional and non-disruptive. When Unix socket support is disabled, the gateway continues to operate exactly as it does today, using TCP/TLS only.

When enabled:

The gateway creates a Unix domain socket at the configured path.

A default socket path (for example, /tmp/osddb.sock) is used if no custom path is provided.

Users may optionally configure a custom socket path, such as:

```json
{
  "UnixSocketEnabled": true,
  "UnixSocketPath": "/tmp/custom_gateway.sock"
}
```

This approach ensures backward compatibility while allowing users to opt in to Unix domain socket support when local IPC is beneficial.

v-sathvik:
- Add unix_socket_enabled and unix_socket_path configuration options
- Implement Unix socket listener alongside TCP in gateway
- Add integration tests for the Unix socket
- Add test helpers for Unix socket configuration

Fixes #400

### DIFF
--- a/pg_documentdb_gw/src/configuration/setup.rs
+++ b/pg_documentdb_gw/src/configuration/setup.rs
@@ -52,8 +52,8 @@ pub struct DocumentDBSetupConfiguration {
     // Unix domain socket configuration
     // If specified with a non-empty path, Unix socket is enabled at that path.
     // If not specified (None), Unix socket is disabled.
-	pub unix_socket_path: Option<String>,
-    
+    pub unix_socket_path: Option<String>,
+
     // Unix socket file permissions (octal format string, e.g., "0660" for owner+group read/write)
     // If not specified, defaults to 0o660
     pub unix_socket_file_permissions: Option<String>,
@@ -65,7 +65,7 @@ impl DocumentDBSetupConfiguration {
         let config: Self = serde_json::from_reader(config_file).map_err(|e| {
             DocumentDBError::internal_error(format!("Failed to parse configuration file: {e}"))
         })?;
-        
+
         // Validate Unix socket path if provided
         if let Some(path) = &config.unix_socket_path {
             if path.trim().is_empty() {
@@ -74,16 +74,16 @@ impl DocumentDBSetupConfiguration {
                 ));
             }
         }
-        
+
         // Validate Unix socket permissions if provided
         if let Some(perm_str) = &config.unix_socket_file_permissions {
             if u32::from_str_radix(perm_str, 8).is_err() {
                 return Err(DocumentDBError::internal_error(
-                    format!("Invalid UnixSocketFilePermissions '{}'. Expected octal format like '0600', '0644'", perm_str)
+                    format!("Invalid UnixSocketFilePermissions '{perm_str}'. Expected octal format like '0600', '0644'")
                 ));
             }
         }
-        
+
         Ok(config)
     }
 }
@@ -170,7 +170,7 @@ impl SetupConfiguration for DocumentDBSetupConfiguration {
     fn unix_socket_path(&self) -> Option<&str> {
         self.unix_socket_path.as_deref()
     }
-    
+
     fn postgres_idle_connection_timeout_minutes(&self) -> u64 {
         self.postgres_idle_connection_timeout_minutes.unwrap_or(5)
     }
@@ -181,8 +181,8 @@ impl SetupConfiguration for DocumentDBSetupConfiguration {
 
     fn unix_socket_file_permissions(&self) -> u32 {
         match &self.unix_socket_file_permissions {
-            None => 0o660,  // Default when not provided
-            Some(perm_str) => u32::from_str_radix(perm_str, 8).unwrap()
+            None => 0o660, // Default when not provided
+            Some(perm_str) => u32::from_str_radix(perm_str, 8).unwrap(),
         }
     }
 }
--- a/pg_documentdb_gw/src/lib.rs
+++ b/pg_documentdb_gw/src/lib.rs
@@ -62,7 +62,7 @@ const STREAM_WRITE_BUFFER_SIZE: usize = 8 * 1024;
 const TLS_PEEK_TIMEOUT_SECS: u64 = 5;
 
 /// Applies configurable permissions to Unix domain socket file.
-/// 
+///
 /// On Unix systems: Sets permissions to the specified octal value
 /// On other platforms: No-op (permissions handled by OS defaults)
 #[cfg(unix)]
@@ -108,16 +108,21 @@ fn create_unix_socket_listener(socket_path: &str, permissions: u32) -> Result<Un
         if e.kind() != std::io::ErrorKind::NotFound {
             tracing::warn!(
                 "Could not remove existing socket file {}: {}.",
-                socket_path, e
+                socket_path,
+                e
             );
         }
     }
-    
+
     let listener = UnixListener::bind(socket_path)?;
-    
+
     apply_socket_permissions(socket_path, permissions)?;
 
-    tracing::info!("Unix socket listener bound to {} with permissions {:o}", socket_path, permissions);
+    tracing::info!(
+        "Unix socket listener bound to {} with permissions {:o}",
+        socket_path,
+        permissions
+    );
     Ok(listener)
 }
 
@@ -164,16 +169,22 @@ where
     ))
     .await?;
 
-    tracing::info!("TCP listener bound to port {}", service_context.setup_configuration().gateway_listen_port());
+    tracing::info!(
+        "TCP listener bound to port {}",
+        service_context.setup_configuration().gateway_listen_port()
+    );
 
-    let unix_listener = if let Some(unix_socket_path) = service_context.setup_configuration().unix_socket_path() {
-        let permissions = service_context.setup_configuration().unix_socket_file_permissions();
-        let unix_listener = create_unix_socket_listener(unix_socket_path, permissions)?;
-        Some(unix_listener)
-    } else {
-        tracing::info!("Unix socket disabled (not configured)");
-        None
-    };
+    let unix_listener =
+        if let Some(unix_socket_path) = service_context.setup_configuration().unix_socket_path() {
+            let permissions = service_context
+                .setup_configuration()
+                .unix_socket_file_permissions();
+            let unix_listener = create_unix_socket_listener(unix_socket_path, permissions)?;
+            Some(unix_listener)
+        } else {
+            tracing::info!("Unix socket disabled (not configured)");
+            None
+        };
 
     // Listen for new tcp and unix socket connections
     loop {
--- a/pg_documentdb_gw/tests/common/mod.rs
+++ b/pg_documentdb_gw/tests/common/mod.rs
@@ -208,19 +208,13 @@ pub fn get_unix_socket_client_custom(path: &str) -> Client {
 }
 
 #[allow(dead_code)]
-pub async fn initialize_with_config_and_unix(
-    path: Option<String>,
-) -> (Client, Option<Client>) {
+pub async fn initialize_with_config_and_unix(path: Option<String>) -> (Client, Option<Client>) {
     let config = setup_configuration_with_unix_socket_custom(path.clone(), None);
     initialize_full(config).await;
-    
+
     let tcp_client = get_client();
-    let unix_client = if let Some(socket_path) = path {
-        Some(get_unix_socket_client_custom(&socket_path))
-    } else {
-        None
-    };
-    
+    let unix_client = path.map(|socket_path| get_unix_socket_client_custom(&socket_path));
+
     (tcp_client, unix_client)
 }
 
--- a/pg_documentdb_gw/tests/unix_socket_tests_disabled.rs
+++ b/pg_documentdb_gw/tests/unix_socket_tests_disabled.rs
@@ -37,17 +37,20 @@ async fn test_unix_socket_disabled() {
 #[tokio::test]
 async fn test_unix_socket_connection_fails_when_disabled() {
     let _ = std::fs::remove_file("/tmp/osddb.sock");
-    
+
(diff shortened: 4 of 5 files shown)


pg_documentdb_gw/src/configuration/setup.rs
```diff
@@ -47,6 +47,11 @@ pub struct DocumentDBSetupConfiguration {
 
     // Runtime configuration
     pub async_runtime_worker_threads: Option<usize>,
+
+    // Unix domain socket configuration
+	pub unix_socket_enabled: Option<bool>,
```
visridha: Curious - do we need both? having the path specified should signal we enable UDS right? is there a reason to have an optional Toggle AND an optional path?

visridha: This would then be similar to the [Postgres configuration for Unix Sockets](https://www.postgresql.org/docs/current/runtime-config-connection.html#GUC-UNIX-SOCKET-DIRECTORIES)

pg_documentdb_gw/src/lib.rs
```diff
@@ -58,7 +58,122 @@ const TCP_KEEPALIVE_INTERVAL_SECS: u64 = 60;
...
+        match self.get_mut() {
+            GatewayStream::Tcp(stream) => Pin::new(stream).poll_shutdown(cx),
+            GatewayStream::Unix(stream) => Pin::new(stream).poll_shutdown(cx),
+        }
+    }
+}
+
+impl std::marker::Unpin for GatewayStream {}
+
+type GwStream = GatewayStream;
```
visridha: Why do we need this? GatewayStream already seems clear enough - do we still want an alias? Could we just replace the callsites with GatewayStream instead?

pg_documentdb_gw/src/lib.rs
```diff
@@ -122,7 +249,31 @@ where
                     }
                 });
             }
+            stream_result = async {
+                match &unix_listener {
```
visridha: nit: Would prefer this to be std::future::ready() in the case of None - since it is not polled in the case of disabled but we don't have to worry about cancelability of the branch.

pg_documentdb_gw/src/lib.rs
```diff
@@ -58,7 +58,122 @@ const TCP_KEEPALIVE_INTERVAL_SECS: u64 = 60;
...
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<std::io::Result<()>> {
+        match self.get_mut() {
+            GatewayStream::Tcp(stream) => Pin::new(stream).poll_shutdown(cx),
+            GatewayStream::Unix(stream) => Pin::new(stream).poll_shutdown(cx),
+        }
+    }
+}
+
+impl std::marker::Unpin for GatewayStream {}
```
AndrewKhoma: can we move this out to some separate path, please?
the reason why I'm asking is that lib.rs is already massive and contain a lot of logic that needs to be split up on some separate path in the service module, so I wouldn't like to give this file even more logic

pg_documentdb_gw/src/lib.rs
```diff
@@ -103,7 +218,19 @@ where
...
+    let unix_listener = if service_context.setup_configuration().unix_socket_enabled() {
+        let unix_socket_path = service_context.setup_configuration().unix_socket_path();
+        let unix_listener = create_unix_socket_listener(&unix_socket_path)?;
+        Some((unix_listener, unix_socket_path))
+    } else {
+        tracing::info!("Unix socket disabled (not configured)");
+        None
+    };
+
+    tracing::info!("TCP listener bound to port {}", service_context.setup_configuration().gateway_listen_port());
```
AndrewKhoma: i would recommend moving this line right after the line 219 (once the tcp_listener) is created

pg_documentdb_gw/src/lib.rs
AndrewKhoma, line 196: should we rename this function to the `handle_tcp_connection` then?

pg_documentdb_gw/tests/unix_socket_tests_default_path.rs
```diff
@@ -0,0 +1,67 @@
...
+
+    // Verify TCP-inserted data can be read via Unix socket
+    let tcp_data = unix_coll.find_one(doc! { "via": "tcp" }).await.unwrap();
+    let tcp_doc = tcp_data.expect("TCP document should exist");
+    assert_eq!(tcp_doc.get_str("via").unwrap(), "tcp");
+
+    // Verify Unix-inserted data can be read via TCP
+    let unix_data = tcp_coll.find_one(doc! { "via": "unix" }).await.unwrap();
+    let unix_doc = unix_data.expect("Unix document should exist");
+    assert_eq!(unix_doc.get_str("via").unwrap(), "unix");
```
AndrewKhoma, line 57: since they're sharing the same db and collection, can you validate also (as a sanity check) that both tcp and unix clients return 2 documents on count?

pg_documentdb_gw/tests/unix_socket_tests_disabled.rs
AndrewKhoma, line 1: i don't see how we validate that the unix socket is not working with this test.
can we create a unix client anyways and verify that the we can't make any requests through it?

pg_documentdb_gw/tests/common/mod.rs
```diff
@@ -179,6 +189,40 @@ pub async fn initialize_with_config(config: DocumentDBSetupConfiguration) -> Cli
...
+pub async fn initialize_with_config_and_unix(
+    enabled: Option<bool>,
+    path: Option<String>,
+) -> (Client, Option<Client>) {
+    let config = setup_configuration_with_unix_socket_custom(enabled, path.clone());
+    initialize_full(config).await;
+    
+    let tcp_client = get_client();
+    let unix_client = if enabled.unwrap_or(false) {
+        let socket_path = path.unwrap_or_else(|| "/tmp/osddb.sock".to_string());
```
AndrewKhoma: shouldn't this be coming from config? since it will be populated anyways as part of `unix_socket_path`

AndrewKhoma (changes requested):
thank you for your changes!

pg_documentdb_gw/src/configuration/setup.rs
v-sathvik: Yes, I considered the PostgreSQL pattern, but went with two fields for better UX.

With just the path, users would need to explicitly write the default path every time (defeating the purpose of defaults). With the enable flag, users can simply set `"UnixSocketEnabled": true` and get the default path automatically, or provide a custom path if needed.

This way:

- `{}` = disabled (backward compatible)
- `{"UnixSocketEnabled": true}` = enabled at default `/tmp/osddb.sock` (convenient)
- `{"UnixSocketEnabled": true, "UnixSocketPath": "/custom"}` = enabled at custom path

Thoughts?

pg_documentdb_gw/src/lib.rs
v-sathvik: Makes sense! I originally kept it as GwStream to minimize the diff when adding Unix socket support, but you're right that it's just extra indirection now.

I'll rename the enum itself to `GwStream` and remove the alias - that way we get rid of the indirection with minimal changes. Let me know what you think?

pg_documentdb_gw/src/configuration/setup.rs
v-sathvik: I've simplified it to just UnixSocketPath following the PostgreSQL pattern
 
 The path's presence now determines enablement:

- `{}` = disabled (backward compatible)
- `{"UnixSocketPath": "/tmp/osddb.sock"}` = enabled at specified path

pg_documentdb_gw/src/configuration/setup.rs
```diff
@@ -137,6 +142,15 @@ impl SetupConfiguration for DocumentDBSetupConfiguration {
...
+    fn unix_socket_enabled(&self) -> bool {
+  		self.unix_socket_enabled.unwrap_or(false)  // Default: disabled
+  	}
+
+    fn unix_socket_path(&self) -> String {
+  		self.unix_socket_path
+  		    .clone()
+  		    .unwrap_or_else(|| "/tmp/osddb.sock".to_string())
+  	}
     fn postgres_idle_connection_timeout_minutes(&self) -> u64 {
```
vtarassov: Nit: Spacing here looks off.

pg_documentdb_gw/src/lib.rs
```diff
@@ -58,7 +58,122 @@ const TCP_KEEPALIVE_INTERVAL_SECS: u64 = 60;
...
+type GwStream = GatewayStream;
+
+/// Applies secure permissions to Unix domain socket file.
+/// 
+/// On Unix systems: Sets permissions to 0660 (owner+group read/write)
+/// On other platforms: No-op (permissions handled by OS defaults)
+#[cfg(unix)]
+fn apply_socket_permissions(path: &str) -> std::io::Result<()> {
+    use std::os::unix::fs::PermissionsExt;
+    let permissions = std::fs::Permissions::from_mode(0o660);
```
vtarassov: Mongo and PG allow this to be configured and we ought to do the same.

pg_documentdb_gw/tests/common/mod.rs
```diff
@@ -179,6 +189,40 @@ pub async fn initialize_with_config(config: DocumentDBSetupConfiguration) -> Cli
...
+
+    let client_options = ClientOptions::builder()
+        .credential(credential)
+        .hosts(vec![ServerAddress::parse(path).unwrap()])
+        .build();
+    Client::with_options(client_options).unwrap()
+}
+
+#[allow(dead_code)]
+pub async fn initialize_with_config_and_unix(
```
vtarassov: I was going to say that we should remove this allow, but it looks like this function is actually not used outside the tests and should be removed.

pg_documentdb_gw/src/lib.rs
```diff
@@ -119,11 +190,35 @@ where
...
+                    )
+                    .await;
+
+                    if let Err(conn_err) = conn_res {
+                        tracing::error!("Failed to accept a Unix socket connection: {conn_err:?}.");
                     }
                 });
             }
             () = token.cancelled() => {
+                if let Some((_, unix_socket_path)) = unix_listener {
```
AndrewKhoma: we will cleanup anyways during the `create_unix_socket_listener`, so is this needed?

AndrewKhoma (commented):
thanks for addressing previous comments! left one last comment

pg_documentdb_gw/src/lib.rs
v-sathvik: Yes, This is not needed. Thank you!

pg_documentdb_gw/src/lib.rs
```diff
@@ -104,7 +164,18 @@ where
     ))
     .await?;
 
-    // Listen for new tcp connections until token is not cancelled
+    tracing::info!("TCP listener bound to port {}", service_context.setup_configuration().gateway_listen_port());
+
+    let unix_listener = if let Some(unix_socket_path) = service_context.setup_configuration().unix_socket_path() {
+        let permissions = service_context.setup_configuration().unix_socket_file_permissions();
+        let unix_listener = create_unix_socket_listener(unix_socket_path, permissions)?;
+        Some((unix_listener, unix_socket_path.to_string()))
```
AndrewKhoma: nit: this can also be cleaned up, since we are not using this part of tuple anywhere

AndrewKhoma: by this I meant `unix_socket_path`

PR #410, opened 2026-01-08
RFC-0009: Pluggable Authentication Architecture

straatsb:
Proposes trait-based authentication providers to enable extensible
authentication mechanisms without modifying core gateway code.

Introduces AuthProvider trait and AuthProviderRegistry to support
independent development of authentication mechanisms by cloud
providers and organizations while maintaining backward compatibility.


rfcs/0009-pluggable-authentication-architecture.md
```diff
@@ -0,0 +1,537 @@
...
+### Non-Goals
+
+This RFC explicitly does NOT:
+
+- Mandate specific authentication mechanisms (providers are pluggable)
+- Change the MongoDB wire protocol or SASL authentication flow
+- Require database schema changes
+- Modify the PostgreSQL authentication system
+- Implement specific cloud provider authentication (those are separate implementations)
+
```
xgerman, line 85: 1) I am a bit confused how we manage users allowed on the database and authentication, e.g. systems like LDAP have groups where you can add useres to the database group, which then drives who has access to the database - in our case who is a user in the database. Some auth providers for LDAP I have seen will after successful authentication *create* the user. Is this something we encourage or support?

2) What is our stance on users wanting to log in via the mongo API but also on postgres with the *same* user. Your approach would preclude that. There is an argument to have the integration point in postgres to take advantage of the many auth plugins. Please eleborate why we shoudl favor your approech.

rfcs/0009-pluggable-authentication-architecture.md
```diff
@@ -0,0 +1,537 @@
...
+### AuthProvider Trait
+
+The core abstraction that all authentication providers must implement. This trait defines the contract for authentication mechanisms.
+
+**Key Methods:**
+
+```rust
+trait AuthProvider: Send + Sync {
+    fn mechanism_name(&self) -> &str;
+    async fn handle_sasl_start(...) -> Result<Response>;
```
AndrewKhoma: nit: #[async_trait] is missing for trait definition

rfcs/0009-pluggable-authentication-architecture.md
straatsb: 1. Great question! I don't see why we couldn't support a "create user on authentication" pattern.

Here's how I'd imagine it working: when you authenticate with LDAP, you provide your credentials and specify LDAP as your mechanism. The gateway, already configured to talk to your LDAP server, reaches out to validate those credentials. Once authentication succeeds, we check if the user exists in Postgres. if not, we create them automatically on that first successful login.

The key thing is that this RFC focuses on making authentication pluggable, while authorization stays Postgres's job. Users need to exist in Postgres for authorization purposes (roles, permissions, etc) and after successful authentication, we map to that Postgres user. Whether that user already exits or gets created on-the-fly is really up to how the specific auth provider is implemented.

So yeah, the architecture definitely supports autocreating users on authentication it's just a matter of implementing that logic in your custom auth provider. Without create-on-first-authentication you have to create/delete each corresponding user manually or have some synchronization between user directories.

2. Good point about dual-access scenarios and leveraging Postgres auth plugins. My approach doesn't stop users from accessing via both APIs whether it works depends on the specific authentication mechanism.

Ex SCRAM-SHA-256, users can authenticate through both paths today: via the MongoDB API where the gateway validates their credentials against Postgres, and directly to Postgres using its native scram-sha-256 authentication. Same user and password stored in Postgres, just different wire protocols doing the validation.

You raise a fair point about Postgres extensions. The pluggable architecture actually supports both approaches. While I'm aiming to have auth logic live in the gateway providers, there's nothing stopping a provider from just decoding the BSON message and delegating to an existing Postgres auth extension. The gateway still needs to handle the MongoDB wire protocol and SASL handshake regardless but once that's parsed a provider could absolutely call into a mature Postgres extension for the actual validation.

So implemeters can choose what makes sense for their use case. They can handle everything in the gateway provider, call Postgres functions for validation (like we already do for scram-sha-256), or delegate to existing Postgres auth extensions. As for cloud providers, this means they can write a lightweight rust provider that translates the MongoDB protocol rather than requiring a full Postgres extension in C. But if a Postgres extension already exists and makes sense to use, the architecture doesn't prevent that.

In my opinion there are a few reasons why handling authn in the gateway layer makes sense. First, the gateway already has to parse the MongoDB wire protocol and SASL handshake. Second, not all mechanisms have mature Postgres extensions (OIDC, cloud-specific auth). Third, I believe its easier for cloud providers to write a Rust trait implementation than a Postgres extension in C. And finally relying on Postgres extensions would require customers to install and configure separate auth plugins alongside the DocumentDB extension.

rfcs/0009-pluggable-authentication-architecture.md
```diff
@@ -0,0 +1,537 @@
...
+
+**Configuration System**
+- Extend setup configuration to support provider configuration
+- Support per-provider configuration sections
+
+---
+
+## Detailed Design
+
+### AuthProvider Trait
```
normtown, line 345: Even though this is cloud-agnostic, it seems that we need consider taking dependencies on external services. Things that come to my mind to discuss: timeout configuration, circuit breaker patterns, and fallback behavior when external auth services are slow/unavailable.

Imagine your gateway is configured with SCRAM (local) and OIDC (calls external IdP). If the OIDC provider's external service becomes slow:
- Auth requests using OIDC start timing out
- Connection threads/tasks get blocked waiting
- Thread pool exhaustion can cascade to affect SCRAM users too
- Gateway becomes unresponsive even though local auth would work fine

It would be good for the RFC to specify per-provider timeouts, a retry or circuit-breaker policy (e.g., token bucket). This prevents wasting resources on calls that will fail, cascading failures from thread exhaustion. You may also want to think about fallback behaviors when a circuit breaker is broken (another mechanism, immediate failure, queue for retry?), and health checks for providers.

rfcs/0009-pluggable-authentication-architecture.md
```diff
@@ -0,0 +1,537 @@
...
+
+```rust
+trait AuthProvider: Send + Sync {
+    fn mechanism_name(&self) -> &str;
+    async fn handle_sasl_start(...) -> Result<Response>;
+    async fn handle_sasl_continue(...) -> Result<Response>;
+    fn supports_continue(&self) -> bool;
+    async fn initialize(&mut self) -> Result<()>;
+    async fn cleanup(...) -> Result<()>;
+}
```
normtown: You have `initialize()` and `cleanup()` methods but you don't explain when they're called or how the lifecycle works. E.g., When is `cleanup()` called? (On connection close (success)? Auth failure? Gateway shutdown? All of the above?)

This matters because providers may hold per-connection state that needs cleanup.

I think you need to describe:
- graceful shutdown (should there be a `shutdown()` separate from a per-connection `cleanup()`?)
- provider deregistration (should you be able to remove a provider at runtime? what happens to in-flight authentications that use that provider?)
- partial initialization failure (`initialize()` is async, so what if it fails halfway through setup? Who cleans up partially initialized resources? Do we need a rollback mechanism?)

normtown: This also matters because resource leaks and undefined behavior can cause issues for running systems. Provider implementers need a clear state machine/graph.

rfcs/0009-pluggable-authentication-architecture.md
```diff
@@ -0,0 +1,537 @@
...
+
+---
+
+## Detailed Design
+
+### AuthProvider Trait
+
+The core abstraction that all authentication providers must implement. This trait defines the contract for authentication mechanisms.
+
+**Key Methods:**
```
normtown: I think you need an error contract. Without it:
- Provider authors will invent inconsistent error handling
- The gateway can't provide consistent error messages to clients
- Debugging auth failures becomes harder (which provider failed and why?)
- One buggy provider can take down the whole gateway

You probably need to specify: a standard error type (how do errors map to wire protocol error codes?), panic handling (should you wrap provider calls in `catch_unwind`, how is a panicking provider isolated from affecting other connections?), hanging/buggy providers (timeouts, eliminating providers that block forever), a way for providers to distinguish retryable vs permanent failures (invalid creds, OIDC server unavailable).

rfcs/0009-pluggable-authentication-architecture.md
```diff
@@ -0,0 +1,537 @@
...
+impl AuthProviderRegistry {
+    async fn register(provider: Box<dyn AuthProvider>) -> Result<()>;
+    async fn get_provider(mechanism: &str) -> Result<Arc<dyn AuthProvider>>;
+    async fn is_supported(mechanism: &str) -> bool;
+}
+```
+
+**Design Characteristics:**
+
+- **Thread-safe**: Uses `RwLock` for concurrent access
```
normtown: I think you need to define the concurrency semantics. Can providers be registered/replaced while auth is in-flight? How does multi-step auth handle provider changes?

```
T0: Connection A starts SCRAM auth (calls get_provider("SCRAM-SHA-256"))
T1: Admin registers new provider (takes write lock)
T2: Connection A calls handle_sasl_continue()
```

- Does Connection A's auth complete successfully?
- Is the provider reference held across the lock, or looked up each call?
- Could the provider be replaced mid-authentication?

Another thought from this example: SCRAM is multi-step: `sasl_start -> sasl_continue`. If the provider is swapped between steps, state from step 1 may be incompatible with the new provider implementation. Should multi-step auth "pin" to the provider instance that started it?

rfcs/0009-pluggable-authentication-architecture.md
```diff
@@ -0,0 +1,537 @@
...
+   │Provider │  │Provider │  │Provider │
+   └─────────┘  └─────────┘  └─────────┘
+```
+
+### Key Benefits
+
+**1. Extensibility Without Core Changes**
+- New authentication mechanisms register themselves with the registry
+- No modifications to routing logic required
+- Providers can be added at compile-time or runtime
```
normtown: What stops an attacker from registering a provider that accepts any credentials? Who can register providers? Is there a signing/verification mechanism for providers?

rfcs/0009-pluggable-authentication-architecture.md
```diff
@@ -0,0 +1,537 @@
...
+
+**2. Provider Independence**
+- Each provider is self-contained with its own dependencies
+- Cloud providers can maintain their authentication code independently
+- Providers can be tested in isolation
```
normtown, line 153: Can a buggy/malicious provider access another provider's state? What prevents a provider from accessing credentials from other mechanisms? Are they sandboxed in any way or do they have full gateway privileges?

rfcs/0009-pluggable-authentication-architecture.md
```diff
@@ -0,0 +1,537 @@
...
+
+---
+
+## Approach
+
+### Solution Overview
+
+Implement a trait-based authentication provider architecture using Rust's type system. Each authentication mechanism becomes a self-contained provider that implements a common `AuthProvider` trait. A central registry routes authentication requests to the appropriate provider based on the SASL mechanism name.
+
+### Core Components
```
normtown, line 94: How are you thinking about audit logging? Auth events are security-sensitive and often required for compliance. Should providers or the gateway be responsible to emit audit logs? How would we enforce that? Should audit logging have a standard format (required fields like user, mechanism, success/fail, timestamp, source IP, etc.).

rfcs/0009-pluggable-authentication-architecture.md
```diff
@@ -0,0 +1,537 @@
...
+
+**1. AuthProvider Trait**
+- Cloud-agnostic interface that all authentication providers implement
+- Supports diverse authentication patterns: credential-based, token-based, certificate-based, multi-step handshakes
+- Providers control their own initialization, state management, and cleanup
```
normtown, line 99: How are you thinking about handling the credentials in memory? E.g., zero-ing after use. Should providers be prohibited from persisting credentials? What about forwarding credentials to external services?

rfcs/0009-pluggable-authentication-architecture.md
```diff
@@ -0,0 +1,537 @@
...
+- Extend setup configuration to support provider configuration
+- Support per-provider configuration sections
+
+---
+
+## Detailed Design
+
+### AuthProvider Trait
+
+The core abstraction that all authentication providers must implement. This trait defines the contract for authentication mechanisms.
```
normtown, line 347: I think we should think about whether we can address some basic timing attacks. Provider implementers may not be security experts.  If the RFC says "providers MUST use constant-time comparison for credential validation" and "providers SHOULD NOT have timing differences between valid/invalid usernames," it sets clear expectations. Without guidance, each provider will handle this differently (or not at all).

I'm thinking about the basics.

E.g., Username enumeration:

```rust
// VULNERABLE: Different code paths = different timing
async fn authenticate(username: &str, password: &str) -> Result<()> {
    let user = db.find_user(username).await?;  // Fast if user doesn't exist

    if user.is_none() {
        return Err(AuthError::InvalidCredentials);  // Returns quickly
    }

    // Only reached if user exists - takes longer
    let hash = expensive_hash(password);
    if hash != user.password_hash {
        return Err(AuthError::InvalidCredentials);
    }
    Ok(())
}
```
An attacker can:
1. Try authenticate("alice", "x") → 50ms response
2. Try authenticate("bob", "x") → 150ms response
3. Conclude: "alice" doesn't exist, "bob" does

Now they know valid usernames to target with password attacks.

Similarly, for password comparison timing:

```rust
// VULNERABLE: Early exit on first mismatch
fn compare(a: &[u8], b: &[u8]) -> bool {
    if a.len() != b.len() { return false; }
    for i in 0..a.len() {
        if a[i] != b[i] { return false; }  // Exits early!
    }
    true
}
```

If first byte matches, there is a slightly longer response. Attacker can guess passwords byte-by-byte.

Mitigations the RFC should mention:

1. Constant-time comparison: compare all bytes regardless of match
```rust
use subtle::ConstantTimeEq;
hash_a.ct_eq(&hash_b)
```
2. Normalize response times - Add artificial delay so all auth attempts take similar time
3. Always do the expensive work - Hash the password even if user doesn't exist
```rust
let user = db.find_user(username).await?;
let hash = expensive_hash(password);  // Always compute

match user {
    Some(u) if hash == u.password_hash => Ok(()),
    _ => Err(AuthError::InvalidCredentials),
}
```

rfcs/0009-pluggable-authentication-architecture.md
```diff
@@ -0,0 +1,537 @@
...
+- Enable/disable providers via configuration
+- Provider-specific configuration (isolated from other providers)
+
+### Architecture Diagram
+
+```
+┌─────────────────────────────────────────────────────────────┐
+│                     MongoDB Client                          │
+└────────────────────┬────────────────────────────────────────┘
+                     │ SASL Request (mechanism: "SCRAM-SHA-256")
```
normtown, line 121: The RFC doesn't cover the `hello` behavior. The gateway need to know which mechanisms to advertise, and a priority order for them.

normtown: One thing to consider is that mechanisms may have conditional availability per user/connection.

rfcs/0009-pluggable-authentication-architecture.md
```diff
@@ -0,0 +1,537 @@
...
+- All existing tests pass
+- No behavior changes in authentication flows
+- Performance metrics unchanged
+- Zero customer impact
+
+**Phase 2: Configuration and Observability**
+
+1. Add provider configuration to setup files
+2. Implement provider enable/disable functionality
+3. Add per-provider metrics and logging
```
normtown: Will providers get a metrics registry passed in? Do they return metrics from a trait method? Do they emit to a global registry?

normtown: Will there be some kind of trace context that gets propagated? Could be as simple a request ID that gets logged and forwarded, or more complicated.

rfcs/0009-pluggable-authentication-architecture.md
```diff
@@ -0,0 +1,537 @@
...
+    
+    fn supports_continue(&self) -> bool {
+        true  // SCRAM requires multi-step handshake
+    }
+    
+    async fn handle_sasl_start(...) -> Result<Response> {
+        // Extract existing SCRAM logic from auth.rs
+        // Parse client-first-message
+        // Generate server-first-message
+        // Store state in connection context
```
normtown: How? Is there an interface for that? Presumably `handle_sasl_continue` needs access to this, i.e., `handle_sasl_continue(state: Self::State, ...)`. Doing typed state is type-safe but it will be complex.

You could also just use opaque bytes, but that lacks type safety. Or the provider could manager their own state through a ConnectionContext interface for that. Something like `fn handle_sasl_start(ctx: &mut ConnectionContext, ...) -> Response;` and then providers call `ctx.set_auth_state(key, value)`. That kind of just pushes the same problems as above but to the `value` parameter.

normtown: Also, should we have size limits? What if a buggy provider stores a gig of state? This could also be a vector for a memory exhaustion attack: start many auth flows but never complete them. When does the state get cleaned up?

rfcs/0009-pluggable-authentication-architecture.md
normtown: Thinking more on this, I'm not clear on what "added at runtime" means. Rust doesn't have a stable ABI AFAIK (I'm not an expert), so I'm not sure how implementations would be loaded while running.

If it's using a C library, then we lose Rust's type safety at the boundary and it's error-prone.

Otherwise, it seems like you mean all providers are compiled in and config enables/disables them. I can't otherwise see how to do dynamic pug-in loading.

I think you need to clarify what "runtime" means. If it's just config toggles, say so. If you're actually planning dynamic loading, then you need to address the ABI stability problem and the security model for loading untrusted code.

rfcs/0009-pluggable-authentication-architecture.md
```diff
@@ -0,0 +1,537 @@
...
+
+**Trait-Based Architecture**
+- Leverages Rust's type system for compile-time safety
+- Enables polymorphism without runtime overhead
+- Provides clear contract for provider implementations
+
+**Registry Pattern**
+- Centralized provider management
+- Dynamic provider registration
+- Supports runtime configuration changes
```
AndrewKhoma: can you describe any scenario in which we will inject a new provider on the flight? I assume that all the auth providers are known at compile time and can't be added say though any API call to the gateway. Would really love to read more on this topic

rfcs/0009-pluggable-authentication-architecture.md
```diff
@@ -0,0 +1,537 @@
...
+
+**Cloud-Agnostic Core**
+- Core authentication logic has no cloud-specific dependencies
+- Providers encapsulate cloud-specific logic
+- Enables multi-cloud and hybrid deployments
```
AndrewKhoma: seems a bit out of scope...

since the current project is already open-sourced and provides the code under MIT and APACHE-2.0 license I don't see how it can be not multi-cloud by definition (or I'm missing something important)

rfcs/0009-pluggable-authentication-architecture.md
```diff
@@ -0,0 +1,537 @@
...
+
+### Tradeoffs
+
+**Abstraction Overhead**
+- *Benefit:* Clean separation of concerns, extensibility
+- *Cost:* Additional indirection through trait dispatch
+- *Mitigation:* Minimal performance impact (single trait call per auth request)
```
AndrewKhoma: nit: there can be a couple auth requests (like SaslStart and SaslContinue) and also since in the current RFC the proposed solution involves dynamic dispatch it will also involve a vtable lookup per every call to the concrete type implementation which is fine and we need to accept this for the better extensibility

rfcs/0009-pluggable-authentication-architecture.md
```diff
@@ -0,0 +1,537 @@
...
+
+**Refactoring Existing Code**
+- *Benefit:* Cleaner architecture, easier to maintain
+- *Cost:* Risk of introducing regressions in existing auth
+- *Mitigation:* Comprehensive test suite, phased rollout, backward compatibility guarantees
```
AndrewKhoma, line 190: i would propose to have the old and new version side by side for the first release and have some GUC to have a fallback mechanics

rfcs/0009-pluggable-authentication-architecture.md
```diff
@@ -0,0 +1,537 @@
...
+- Refactor `auth.rs` to use registry pattern
+- Add provider trait and registry modules
+
+**ServiceContext**
+- Add `auth_provider_registry` field
+- Initialize registry with configured providers
+- Provide registry access to connection handlers
+
+**Configuration System**
+- Extend setup configuration to support provider configuration
```
AndrewKhoma, line 260: maybe each provider should have their own config for flexibility?
i had this idea in my mental backlog for quite some time: we need to refactor all configs in some config registry as you're describing it for auth providers, so we can pull needed configs from either setup config or from in-memory config or from postgres, but also it can be extended by other provider configs

rfcs/0009-pluggable-authentication-architecture.md
```diff
@@ -0,0 +1,537 @@
...
+    fn supports_continue(&self) -> bool;
+    async fn initialize(&mut self) -> Result<()>;
+    async fn cleanup(...) -> Result<()>;
+}
+```
+
+**Design Characteristics:**
+
+- **Cloud-agnostic**: No dependencies on specific cloud providers
+- **Flexible**: Supports credential-based, token-based, certificate-based, and multi-step authentication patterns
```
AndrewKhoma: nit: don't repeat yourself? it was outlined already in the ##approach section

rfcs/0009-pluggable-authentication-architecture.md
AndrewKhoma: also, if concurrent updates in-flight are supported, then should the providers use DashMap instead of HashMap?

also I'm a bit lost here, what exactly this RwLock will be holding? the auth registry provider?

rfcs/0009-pluggable-authentication-architecture.md
```diff
@@ -0,0 +1,537 @@
...
+- **O(1) lookup**: HashMap provides constant-time provider lookup
+- **Duplicate prevention**: Rejects registration of duplicate mechanism names
+- **Initialization**: Calls provider `initialize()` during registration
+
+**Registry Behavior:**
+
+1. **Registration**: Providers register themselves with their mechanism name
+2. **Lookup**: Authentication requests query the registry by mechanism name
+3. **Validation**: Registry ensures no duplicate mechanisms are registered
+4. **Error handling**: Returns clear errors for unsupported mechanisms
```
AndrewKhoma, line 547: purely for better understanding, can it be described as sequence diagram instead of text?

rfcs/0009-pluggable-authentication-architecture.md
```diff
@@ -0,0 +1,537 @@
...
+```
+
+**Proposed Implementation (Registry-Based):**
+
+```rust
+async fn handle_sasl_start(...) -> Result<Response> {
+    let mechanism = request.document().get_str("mechanism")?;
+    let registry = connection_context.service_context.auth_provider_registry();
+    let provider = registry.get_provider(mechanism).await?;
+    provider.handle_sasl_start(connection_context, request).await
```
AndrewKhoma, line 613: which means that each provider should have an implementation for handle sasl start? what if someone wants to implement kerberos auth which involves (if i recall correctly) 3 round trips

maybe provider should simply take a `request.document()` and call some `authenticate` method instead?

rfcs/0009-pluggable-authentication-architecture.md
```diff
@@ -0,0 +1,537 @@
...
+        // Parse client-first-message
+        // Generate server-first-message
+        // Store state in connection context
+    }
+    
+    async fn handle_sasl_continue(...) -> Result<Response> {
+        // Extract existing SCRAM continue logic
+        // Verify client proof
+        // Complete authentication
+    }
```
AndrewKhoma: so yeah, I'm coming back to my `authenticate` method logic, i feel like if we will provide such facade method that knows depending on the inner state and the document itself knows what to call next

rfcs/0009-pluggable-authentication-architecture.md
```diff
@@ -0,0 +1,537 @@
...
+}
+```
+
+**Refactoring Strategy:**
+
+1. **Extract logic**: Move existing authentication code into provider implementations
```
AndrewKhoma, line 783: should we build a new logic alongside existing one and the deprecate and old implementation?

rfcs/0009-pluggable-authentication-architecture.md
```diff
@@ -0,0 +1,537 @@
...
+2. **Preserve behavior**: Ensure refactored providers produce identical results
+3. **Maintain state**: Use existing connection context for state management
+4. **No API changes**: External behavior remains unchanged
+
+**Provider Examples:**
+
+- **ScramProvider**: Multi-step credential-based authentication (existing)
```
AndrewKhoma, line 790: what about PlainProvider?

rfcs/0009-pluggable-authentication-architecture.md
AndrewKhoma: what per-provider metrics can you name? it's definitely good to have a metrics in terms of successful/unsuccessful/total auth requests, but that's a generic logic, not a per-provider one

rfcs/0009-pluggable-authentication-architecture.md
```diff
@@ -0,0 +1,537 @@
...
+
+**Phase 3: Extensibility (Future)**
+
+1. Cloud providers can add authentication mechanisms
+2. Organizations can implement custom providers
+3. Community contributions for additional methods
+
+**Rollback Strategy:**
+
+- Feature flag to switch between old and new implementation
```
AndrewKhoma: nit: GUC?

rfcs/0009-pluggable-authentication-architecture.md
```diff
@@ -0,0 +1,537 @@
...
+- Comprehensive monitoring during rollout
+- Automated rollback on error rate increase
+- Keep original implementation available for quick revert
+
+### Documentation Updates
+
+**User-Facing Documentation**
+
+- Overview of supported authentication mechanisms
```
AndrewKhoma, line 1065: who's providing this type of documentation? @abinav2307 is it you or Sudhanshu?

AndrewKhoma (commented):
great RFC! looking forward to collaborate on gw changes!

rfcs/0009-pluggable-authentication-architecture.md
straatsb: Good catch, the 'feature flag' language doesn't make sense here. This is a refactor, not a parallel implementation. I've changed this section into Deployment Considerations

rfcs/0009-pluggable-authentication-architecture.md
straatsb: Most metrics would be generic (success/failure/latency) but tagged with the provider name for filtering. Some provider-specific ones could include: OIDC token expiry events, SCRAM handshake step timing, or future IAM provider's AWS API call latency. But you're right that the core metrics are generic. I've clarified this to mainly be about tagging generic metrics by provider rather than entirely different metric types per provider.

rfcs/0009-pluggable-authentication-architecture.md
straatsb: Yea, PLAIN would work fine with this architecture. It's not currently supported (only SCRAM-SHA-256 and MONGODB-OIDC), but the trait supports both single-step and multi-step authentication patterns. I've updated the RFC to include PLAIN and other future mechanims in the provider examples

rfcs/0009-pluggable-authentication-architecture.md
straatsb: I'm planning to do a clean refactor rather than running both implementations in parallel. The existing auth code will be extracted into providers (ScramProvider, OidcProvider) and the old monolithic handler will be replaced entirely. The refactoring strategy is to preserve behavior while restructuring the code, not to maintain two separate implementations. If issues arise during deployment, users can roll back to the previous release. Do you think thats safe enough or would you prefer running both in parallel?

rfcs/0009-pluggable-authentication-architecture.md
straatsb: Fixed

rfcs/0009-pluggable-authentication-architecture.md
straatsb: You're right providers would be known at compile time. I was doing a little too much when I initially wrote this RFC looking into runtime changes and the first draft had a bunch of references to it and this one was missed when I removed them.

All auth providers would be built at compile time and users can enable/disable them in a config at initial runtime

straatsb: I've updated the RFC to reflect this clearly

rfcs/0009-pluggable-authentication-architecture.md
straatsb: You're right, i intended this section to mean it remains possible to deploy on different cloud providers, this kinda makes things not clear. I've changed the workding to be more clear about how the code should be cloud agnostic and allow for different cloud providers to create and deploy their own auths easily

rfcs/0009-pluggable-authentication-architecture.md
straatsb: true, the two part auth implementations would involve more than a single call. I've removed the single trait call portion

Yea a vtable lookup is needed, i'll keep it as minimal perf impact but i would expect it to be very minimal

rfcs/0009-pluggable-authentication-architecture.md
straatsb: I read through your comments on this and have been going a bit back and forth but I think I've landed on I think its better to refactor this into flexible auth providers fully without providing the ability to switch between the old monolithic one and the pluggable versions via a GUC.

My thought is this is mainly a refactor for phase 1 where the business logic largely stays the same for each provider I just move each one into a different file and create an auth lookup mapping on user authentication. I also worry about the deprecation timeline and keeping the monolithic code around for a significant amount of time and keeping them both in sync if there are future auth changes during the transition time.

I'm open to suggestions if you think differently, I'm also not sure if theres any precedent for this kind of thing or the common approach is to provide the old and new options under a GUC

rfcs/0009-pluggable-authentication-architecture.md
straatsb: ```maybe each provider should have their own config for flexibility?```
Yes, they can provide additional configs under each auth section in the JSON config file, not just `enabled`. That being said, are you saying we should have separate files or some other way to set configs for auths?

``` i had this idea in my mental backlog for quite some time: we need to refactor all configs in some config registry as you're describing it for auth providers, so we can pull needed configs from either setup config or from in-memory config or from postgres, but also it can be extended by other provider configs```
I like this idea, it's cumbersome to set configs in different places and having to read them all in code. I don't think a full config registry fits into this RFC, but what do you think about designing the provider config access through an abstraction (like a trait) rather than directly accessing the JSON structs? That way, providers would call something like `config.get("auth.scram.iterations")` instead of `setup_config.auth.providers.scram.iterations`. For now it would just read from the JSON file, but later we could swap in a config registry that pulls from multiple sources without changing provider code

rfcs/0009-pluggable-authentication-architecture.md
straatsb: ack: i removed this part and made sure the points were covered by the approach section fully

rfcs/0009-pluggable-authentication-architecture.md
AndrewKhoma: it's rather our policy that in case we are changing things significantly during the release we introduce GUC and turn it off, then we turn it on some selective accounts (usually they are our test-in-prod) accounts and see if no regressions are there, then the next release is GUC is turned on and we are monitoring the whole fleet and then on the release X+2 we deprecate the code and drop the GUC.

I'm very in favor of this refactoring, so what I would propose is to currently just rename/move out the existing auth.rs and do the refactoring POC and let's see how hard will it be to connect it back. ironically, i think that the current spaghetti should be plug-n-play in our case, but I don't want you to be blocked by the previous decisions

rfcs/0009-pluggable-authentication-architecture.md
AndrewKhoma: >That being said, are you saying we should have separate files or some other way to set configs for auths?

we can start simple - subsection of the setup config json, but that thing will be hard to maintain properly; having separate files - not sure it's will be easy to manage them for deployments and packaging; i was thinking more in terms of GUCs or at least the ability to load them in-memory from provider

>what do you think about designing the provider config access through an abstraction (like a trait) rather than directly accessing the JSON structs?

yup, that was my plan, let me share it here:
image
so at most you need to hardcode the name of the config and then it would be a problem of config_store to lookup from which registered config collector this config is coming

>later we could swap in a config registry that pulls from multiple sources without changing provider code

been there, done it :)
image

rfcs/0009-pluggable-authentication-architecture.md
straatsb: Okay makes sense, I'm happy to follow the policy. I've kept this RFC to describe the end state and deployment goals, but the usual GUC approach (disabled by default, then enabled, then enabled and remove old code) works for me when it comes to deployment.

rfcs/0009-pluggable-authentication-architecture.md
straatsb: I see what you're saying but as far as I've seen from looking around all modern MongoDB authentication mechanisms use the SASL protocol with exactly two commands: saslStart and saslContinue. I could have missed one so lemme know if im wrong.

Multi-round authentication (like Kerberos/GSSAPI with 3+ round trips) is handled by calling `saslContinue` multiple times until the server responds with done: true.

Having separate `handle_sasl_start` and `handle_sasl_continue` methods provides type safety at the protocol level, `saslStart` requires a mechanism name while `saslContinue` requires a conversation id. A generic authenticate request method would lose this type safety and might make it less clear which phase of the sasl conversation were in.

rfcs/0009-pluggable-authentication-architecture.md
straatsb: Answered here https://github.com/documentdb/documentdb/pull/410#discussion_r2755959676

I think I'm still leaning towards keeping saslstart and saslcontinue methods but open to changing that if you feel strngly

AndrewKhoma: it make sense to do multiple sasl_continue calls, thanks for researching!

rfcs/0009-pluggable-authentication-architecture.md
straatsb: Makes sense, i like that. I've refactored the RFC into two main phases for implementation. Phase 1 is largely a refactor of the existing code into a pluggable architecture without changing the majority of business logic. Phase 2 is to add new features such as the things you mentioned above, metrics, additional logging, etc

straatsb: Phase 3 would be contributors adding new auth mechanisms but thats not apart of this RFCs scope

rfcs/0009-pluggable-authentication-architecture.md
straatsb: ack makes sense. I've clarified the lifecycle in the trait definition:

`initialize()` - Called once at startup. Failure = provider not registered.
`shutdown()` - Called once at gateway shutdown (30s timeout).
`on_connection_close()` - Called when connection closes (optional, default no-op).

Each trait method now has comments explaining when it's called and what it's responsible for. Added lifecycle diagram and examples (SCRAM vs OIDC) to show the patterns

rfcs/0009-pluggable-authentication-architecture.md
straatsb: Good point. I've added an "Error Handling" section to the AuthProvider Trait that covers:
- Standard error types (using existing DocumentDBError)
- Phase 2 resilience: panic handling, timeout enforcement, circuit breaker for retryability

rfcs/0009-pluggable-authentication-architecture.md
straatsb: I have removed all references to runtime changing of auth mechanisms. I was initially playing around with the idea and some of those ideas got left in the RFC. it should be all removed now. I don't think the benefit outweighs the complexity and I don't imagine changing auths during runtime is that desirable of a feature, at least for now

rfcs/0009-pluggable-authentication-architecture.md
straatsb: Removed dynamic reloading of configs and auth mechanism, this was something I was cooking up but I've decided to remove it

rfcs/0009-pluggable-authentication-architecture.md
straatsb: Added a Security Model section to the RFC. Currently providers are trusted code with full gateway privileges,only first-party and vetted cloud provider implementations should be supported. State isolation is enforced through code review rather than sandboxing.

Do you think this is suffiecient or should some type of sandboxxced or enforced seperation be mandatory?

rfcs/0009-pluggable-authentication-architecture.md
straatsb: Some auth logging exists today but it's not comprehensive. I've added audit logging to Phase 1 of the RFC. I'm thinking the auth handler (not individual providers) will log all authentication attempts and results with standard fields for compliance. Anhy auth specific audit log can be added to the specific auth handler itself

rfcs/0009-pluggable-authentication-architecture.md
straatsb: Added credential handling guidelines to the Security Model section. Providers must not persist credentials to disk, but may forward them to external services for validation.

Im thinking since providers are trusted code that passes code review, credential handling is the responsibility of the provider implementer rather than being technically enforced. This should follow the same trust model as the rest of the architecture

rfcs/0009-pluggable-authentication-architecture.md
straatsb: Good point on timing attacks. I've added timing attack mitigation to Phase 3. The auth handler will enforce a configurable minimum authentication time per provider to prevent username enumeration and timing-based attacks. This is enforced at the architecture level rather than relying on each provider implementation, and can be dynamically adjusted based on observed patterns.

I did go back and forth on this if it was in scope so I'm curious to know if you feel strongly about this being in this RFC or in the scope for phase 3. I'm also curious to know if a generic timing attack safety net for all auths is better or implementing one on a per authentication mechanism is better since it gives the auth mechanism more control

rfcs/0009-pluggable-authentication-architecture.md
straatsb: ack, added details for dynamic auth mechanism lookup to phase 1 and added per user auth mechanism lookup to phase 3 since its not currently supported

rfcs/0009-pluggable-authentication-architecture.md
straatsb: Metrics will be emitted by the auth handler and registry (not individual providers) to ensure consistency. The gateway already has a connection_id (UUID) in ConnectionContext that's used for trace correlation in logs and providers have access to this for their own logging if needed. I've clarified this in the Phase 2 metrics section.

rfcs/0009-pluggable-authentication-architecture.md
straatsb: Good point on state management. I've added a State Management section to the RFC. Providers store per-connection state in ConnectionContext (keyed by provider name for isolation). State has size limits to prevent memory exhaustion attacks and is automatically cleaned up on connection close or auth timeout. This follows the existing pattern but makes it explicit and adds protections.

I have no data or strong thoughts on state size limits so I've kept it open for now but interested to hear your thoughts

rfcs/0009-pluggable-authentication-architecture.md
straatsb: Edit: I've changed the phases since phase 3 is implicit in that of course contributors can add new auth mechanisms once the refactor is done.

Phase 1: main refactor into auth provider
Phase 2: basic timeouts, configurations, and more logging
Phase 3: Advanced security as mentioned in other parts of this PR

rfcs/0009-pluggable-authentication-architecture.md
straatsb: Update: I've left the 3 part release strategy as an open question. I think I will go forward with releasing it under a GUC unless I've gotten an explicit go ahead to skip that process since its the usual process and I'm just a chill new guy to this repo

rfcs/0009-pluggable-authentication-architecture.md
straatsb: Added a diagram

rfcs/0009-pluggable-authentication-architecture.md
straatsb: Honestly, not totally sure of the documentation process for this repo myself. Is there documentation about expected documentation changes for new features?

PR #411, opened 2026-01-09
BUG Fix: CurrentOp: Changes to ensure we pass correct params from gateway for currentOp

ospooja:
## Corrected Summary

CurrentOp fails when called through the gateway (pymongo) due to a parameter mismatch. The backend function expects 1 BSON document, but the gateway passes 3 separate parameters, causing a PostgreSQL "function does not exist" error.

Backend SQL function:
```sql
CREATE OR REPLACE FUNCTION documentdb_api.current_op_command(
    p_spec documentdb_core.bson,  -- ← Only 1 parameter!
    OUT document documentdb_core.bson
)
```

**But gateway passes 3 params:**
```rust
async fn execute_current_op(...) -> Result<Response> {
    let current_op_rows = self
        .pull_connection(connection_context)
        .await?
        .query(
            connection_context.service_context.query_catalog().current_op(),
            &[Type::BYTEA, Type::BOOL, Type::BOOL],  // ← Expects 3 params
            &[&PgDocument(filter), &all, &own_ops],  // ← Passes 3 params
```

**Query catalog:**
```rust
current_op: "SELECT documentdb_api.current_op($1, $2, $3)".to_string(),
```

**Made changes to include parameters in the BSON document and pass 1 parameter as expected per the function definition.**

### DIFF
--- a/pg_documentdb_gw/src/postgres/data_client.rs
+++ b/pg_documentdb_gw/src/postgres/data_client.rs
@@ -222,7 +222,7 @@ pub trait PgDataClient: Send + Sync {
 
     async fn execute_current_op(
         &self,
-        request_context: &mut RequestContext<'_>,
+        request_context: &RequestContext<'_>,
         connection_context: &ConnectionContext,
     ) -> Result<Response>;
 
--- a/pg_documentdb_gw/src/postgres/documentdb_data_client.rs
+++ b/pg_documentdb_gw/src/postgres/documentdb_data_client.rs
@@ -805,7 +805,7 @@ impl PgDataClient for DocumentDBDataClient {
 
     async fn execute_current_op(
         &self,
-        request_context: &mut RequestContext<'_>,
+        request_context: &RequestContext<'_>,
         connection_context: &ConnectionContext,
     ) -> Result<Response> {
         let (request, request_info, request_tracker) = request_context.get_components();
--- a/pg_documentdb_gw/src/processor/process.rs
+++ b/pg_documentdb_gw/src/processor/process.rs
@@ -33,7 +33,7 @@ enum Retry {
 }
 
 pub async fn process_request(
-    request_context: &mut RequestContext<'_>,
+    request_context: &RequestContext<'_>,
     connection_context: &mut ConnectionContext,
     pg_data_client: impl PgDataClient,
 ) -> Result<Response> {
--- a/pg_documentdb_gw/src/processor/data_management.rs
+++ b/pg_documentdb_gw/src/processor/data_management.rs
@@ -229,7 +229,7 @@ pub async fn process_db_stats(
 }
 
 pub async fn process_current_op(
-    request_context: &mut RequestContext<'_>,
+    request_context: &RequestContext<'_>,
     connection_context: &ConnectionContext,
     pg_data_client: &impl PgDataClient,
 ) -> Result<Response> {


pg_documentdb_gw/src/postgres/documentdb_data_client.rs
```diff
@@ -787,8 +793,8 @@ impl PgDataClient for DocumentDBDataClient {
                     .service_context
                     .query_catalog()
                     .current_op(),
-                &[Type::BYTEA, Type::BOOL, Type::BOOL],
-                &[&PgDocument(filter), &all, &own_ops],
+                &[Type::BYTEA],
+                &[&PgDocument(&combined_doc)],
```
AndrewKhoma: from what I see it should be a request_context.payload since we do the extraction from it here: https://github.com/documentdb/documentdb/blob/2b533c354b968fd9416909a7ce8c8c8cf3fa17e4/pg_documentdb_gw/src/processor/data_management.rs#L215C1-L216C1

so both `filter`, `all` and `own_ops` can be ignored for this newer implementation of current_op

AndrewKhoma: you can see that the postgres is doing all the needed parsing of the bson doc passed to it here:
https://github.com/documentdb/documentdb/blob/2b533c354b968fd9416909a7ce8c8c8cf3fa17e4/pg_documentdb/src/commands/current_op.c#L1714

ospooja: Makes sense, thanks for sharing all the references. I can see that postgres parses through everything so we can just forward the payload.

ospooja: I have pushed the latest changes, @AndrewKhoma could you please take a look when you get a chance.

AndrewKhoma: sorry, forgot to sign-off

AndrewKhoma:
overall LGTM, but can we also cover it with some testing to prevent this situation in the future?

pg_documentdb_gw/src/postgres/query_catalog.rs
```diff
@@ -475,7 +475,7 @@ pub fn create_query_catalog() -> QueryCatalog {
             count_query: "SELECT document FROM documentdb_api.count_query($1, $2)".to_string(),
             coll_stats: "SELECT documentdb_api.coll_stats($1, $2, $3)".to_string(),
             db_stats: "SELECT documentdb_api.db_stats($1, $2, $3)".to_string(),
-            current_op: "SELECT documentdb_api.current_op($1, $2, $3)".to_string(),
+            current_op: "SELECT documentdb_api.current_op($1)".to_string(),
```
AndrewKhoma: while testing it out locally I found that the actual name of the function is `current_op_command`, you can see the definition here: https://github.com/documentdb/documentdb/blob/main/pg_documentdb/sql/udfs/commands_diagnostic/current_op--latest.sql

ospooja: Done!

ospooja: @AndrewKhoma Could you please take a look at the latest changes.

pg_documentdb_gw/tests/currentop_tests.rs
```diff
@@ -0,0 +1,68 @@
...
+    let result = db.run_command(doc! {"currentOp": 1}).await.unwrap();
+    
+    // Verify currentOp returns expected structure
+    assert!(result.contains_key("ok"), "Response should have 'ok' field");
+    assert!(result.contains_key("inprog"), "Response should have 'inprog' field");
+    
+    let ok_value = result.get_f64("ok").unwrap();
+    assert_eq!(ok_value, 1.0, "Expected ok to be 1.0");
+    
+    let inprog = result.get_array("inprog").unwrap();
```
AndrewKhoma: 1. i see that the formatting is off on this test (see https://github.com/documentdb/documentdb/actions/runs/21080130781/job/60742638292) or run `cargo fmt --all --check` yourself
2. i'm not sure how this test is valid, since we don't have any other command started in the background, so I would expect the inprog to be empty, maybe in a separate thread before running currentOp run a transaction with some finds, then run the currentOp, then finish that thread

ospooja: Yes, the format fix is done. 
For the test, since initialize starts creating/dropping db etc, and/or gateway operations, I could always see currentOp return some operations. Even currentOp command should create an operation, so I didn't add any explicitly. It makes sense to add an operation we can track for testing, I added a complex query so we can capture it in the test.

pg_documentdb_gw/tests/currentop_tests.rs
```diff
@@ -0,0 +1,132 @@
...
+    // Give operations time to start
+    tokio::time::sleep(Duration::from_millis(50)).await;
+
+    // Call currentOp to capture the operations
+    let result = db
+        .run_command(doc! {"currentOp": 1, "$all": true})
+        .await
+        .unwrap();
+
+    println!("\n=== currentOp result ===");
```
AndrewKhoma: thanks for updating testing!
nit: can we drop these prints, looks like they were part of the debugging

pg_documentdb_gw/tests/currentop_tests.rs
AndrewKhoma: thank you for the update!

pg_documentdb_gw/tests/currentop_tests.rs
ospooja: Done!

visridha (approved):

visridha:
Looks like the change fails cargo lints

pg_documentdb_gw/src/postgres/data_client.rs
```diff
@@ -222,10 +222,7 @@ pub trait PgDataClient: Send + Sync {
 
     async fn execute_current_op(
         &self,
-        request_context: &RequestContext<'_>,
-        filter: &RawDocumentBuf,
-        all: bool,
-        own_ops: bool,
+        request_context: &mut RequestContext<'_>,
```
AndrewKhoma: mutable borrow is excessive here

pg_documentdb_gw/src/processor/process.rs
```diff
@@ -33,7 +33,7 @@ enum Retry {
 }
 
 pub async fn process_request(
-    request_context: &RequestContext<'_>,
+    request_context: &mut RequestContext<'_>,
```
AndrewKhoma: same

pg_documentdb_gw/src/postgres/data_client.rs
ospooja: Makes sense!

pg_documentdb_gw/src/processor/process.rs
ospooja: Makes sense, made the changes to remove where mut not required

PR #440, opened 2026-02-02
Adding RFC draft proposal for Standardization of Gateway / Extension …

seoc01:
Two main components are:
* All commands implemented in the database extension are called from the gateway passing the command document, and optionally any additional BSON sequences (OP_MSG Section Type 1)
* The Gateway only calls one database extension function per command.


rfcs/0008-gateway-extension-communication-standard.md
```diff
@@ -0,0 +1,262 @@
+|rfc	|title	|status	|owner	|issue	|discussion	|version-target	|implementations	|
```
AndrewKhoma: nit: can you follow the template layout https://raw.githubusercontent.com/documentdb/documentdb/refs/heads/main/rfcs/0000-template.md

e.g.
```markdown
---
rfc: 0000
title: "RFC Title"
status: Draft
owner: "@github-username"
issue: "https://github.com/documentdb/documentdb/issues/XXX"
discussion: "https://github.com/documentdb/documentdb/discussions/XXX"
version-target: 1.0
implementations:
  - "https://github.com/documentdb/documentdb/pull/XXX"
---
```

AndrewKhoma: and looks like 0008 is taken, please, take a 0010

rfcs/0008-gateway-extension-communication-standard.md
```diff
@@ -0,0 +1,262 @@
...
+
+The modifies the command to add the “[global command arguments](https://github.com/mongodb/specifications/blob/master/source/message/OP_MSG.md#global-command-arguments)” (generally $db and $readPreference) to the BSON message, and convert them into the [OP_MSG](https://github.com/mongodb/specifications/blob/master/source/message/OP_MSG.md) format to send over the wire. The data within the OP_MSG is divided into sections of type 0 (main command) and type 1 (document sequences). These sections can come in any order in the OP_MSG.
+
+### Main Command: Type 0
+
+The OP_MSG must send exactly one type 0 section that contains the main body of the command, it is sent as a single BSON object, it must contain all arguments for the command and the global command arguments, except for arguments that are explicitly allowed in the Type 1 sequence based on the command. These arguments are:
+
```
AndrewKhoma, line 25: this feels incomplete, are you talking about the lsid, db and other args?

rfcs/0008-gateway-extension-communication-standard.md
```diff
@@ -0,0 +1,262 @@
...
+|---	|---	|---	|---	|---	|---	|---	|---	|
+
+https://github.com/documentdb/documentdb/pull/XXX
+
+
+# RFC-XXXX: Standardization of Gateway / Extension Interaction
+
+## Background Information
+
+The DocumentDB Gateway is responsible for working as a layer to translate message from the MongoDB clients applications are using, to the SQL functions defined in the Postgres Extension that implement the command. The MongoDB drivers are generally thought of as sending commands in the form of a single BSON message, like would be sent when using the [`runCommand` feature from a driver](https://github.com/mongodb/specifications/blob/master/source/run-command/run-command.md).  In reality they send a structured datagram that is composed of multiple document sections.
```
AndrewKhoma, line 18: i've checked the licensing of this repo https://github.com/mongodb/specifications/blob/master/LICENSE.md
and looks like we can't cite them unless we're given a direct permission. please, do correct me if I'm wrong since I'm not good at the licensing, but I would be highly cautious when it comes to mongodb licensing, cause [they like to sue](https://dockets.justia.com/docket/delaware/dedce/1:2025cv00641/89247)

rfcs/0008-gateway-extension-communication-standard.md
```diff
@@ -0,0 +1,262 @@
...
+bson   document = {
+           'insert': 'city', 
+           '$db': 'database_name'
+           '$readPreference': {...}
+        }
+```
+
+## Problem 
+
+The purpose of the Gateway is to inspect the incoming message from the client, and call the correct database command to complete the request. Currently the Gateway has different logic depending on which command the user is running, and in some cases will parse the command BSON and extract relevant fields to pass to the API defined in the extension, and in other cases it will pass the command directly.
```
AndrewKhoma, line 127: to not mix the causalities here, gw doesn't extracts because it's simply wants so, in some cases the initial API asked to parsed the document, then API evolved, but it's never the gw that dictates the rules, but the extension

rfcs/0008-gateway-extension-communication-standard.md
```diff
@@ -0,0 +1,262 @@
...
+        }
+```
+
+## Problem 
+
+The purpose of the Gateway is to inspect the incoming message from the client, and call the correct database command to complete the request. Currently the Gateway has different logic depending on which command the user is running, and in some cases will parse the command BSON and extract relevant fields to pass to the API defined in the extension, and in other cases it will pass the command directly. 
+
+This adds additional overhead to the Gateway, and makes any consistent handling of commands in the extension difficult, for example adding a function that runs before every command. We need this functionality to support features like the $currentOp aggregation stage, where the extension will need to be able to filter / group across the commands the user is running. 
+
+With the current approach each command would need a custom implementation to be able to re-construct the original command, and some data for some commands, such as lsid, is currently unavailable.
```
AndrewKhoma: that's not true, lsid is parsed here: https://github.com/documentdb/documentdb/blob/9a6ab6c167b1792b039fed7b22ef694b35403125/pg_documentdb_gw/src/requests/mod.rs#L374C1-L375C1
and is available at request_info struct as `session_id` field. the only way the session_id may not be populated is because driver didn't send it

rfcs/0008-gateway-extension-communication-standard.md
```diff
@@ -0,0 +1,262 @@
...
+
+As the document sequences are always optional, any user that does not interact with the gateway (e.g. someone executing the SQL commands directly) can always use the BSON only command operation without needing to construct document sequences if they do not wish to. 
+
+### Alternate Approach
+
+An alternate approach was considered where the gateway took the contents of the document sequences, and inserted them into the command bson as an array instead of sending them as separate components. This would simplify the implementation within the Postgres extension as all commands would have a single request, and they would not need to be aware of how the OP_MSG format splits messages. This would require that the incoming BSON be allowed to exceed 16Mb, which would be acceptable since the 16Mb limit is Mongo defined, and a BSON by itself can be up to 4Gb. 
+
+This is not preferred because it would require additional work at the Gateway layer, where the Gateway may need to create a new several Mb message before passing to the extension, and would prevent the extension from being able to validate the command message is under 16Mb without, .
```
AndrewKhoma: the sentence feels unfinished here

rfcs/0008-gateway-extension-communication-standard.md
```diff
@@ -0,0 +1,262 @@
...
+
+### Example
+
+A find request in the form:
+
+```
+{'find': 'coll', 'filter': {'a':1}, '$db': 'foo'}
+```
+
+Must call a **documentdb_api** function with only a single required argument
```
AndrewKhoma: this is a case for today's gw where each user is having their dedicated connection pool to postgres. We're working on creating a shared pool to cap the total number of connections to postgres from gateway and in this case we would need to pass additional data like user id as a separate parameter to the wrapper extension function, that would authenticate as a user on postgres side and then perform the find operation.

what's the motivation behind single argument requirement here?

AndrewKhoma (changes requested):
thank you for your RFC and looking forward to collaborate on gw part! left some comments to polish this RFC

rfcs/0008-gateway-extension-communication-standard.md
seoc01: Sure, do we have a system to know which ones are taken?

rfcs/0008-gateway-extension-communication-standard.md
seoc01: (also not a lawyer) My understanding is that license applies to copying / distributing the content of the repository. Referencing it / linking to it would not cause any issues. Do you know anyone we could confirm this with?

rfcs/0008-gateway-extension-communication-standard.md
seoc01: I mean everything required for the command, (e.g. command name, collection, $db, any arguments) except (optionally) what is referenced below. If that answers your question I can make the section clearer

rfcs/0008-gateway-extension-communication-standard.md
seoc01: It's good context to have. The purpose of the RFC is to set rules for both the Gateway and the Extension, so the gateway does not have to support custom arguments per command going forward.

rfcs/0008-gateway-extension-communication-standard.md
seoc01: It's parsed in the gateway, but for operations like 'list_databases' (https://github.com/documentdb/documentdb/blob/main/pg_documentdb_gw/src/postgres/documentdb_data_client.rs#L584) I don't think have access to it within the extension

rfcs/0008-gateway-extension-communication-standard.md
seoc01: Yup, missed that. It was supposed to say "and would prevent the extension from being able to validate the command message is under 16Mb without additional knowledge on how the caller constructed the message.

rfcs/0008-gateway-extension-communication-standard.md
seoc01: The goal is to have a single database call that "does work". The specific use case is to allow the postgres extension to correctly server a query with the $currentOp aggregation, which requires that all the main work is done in a single Postgres request, and the extension has the complete client request. Wrapping the request would not break that. I'll see if I can reword it to make is more generic

seoc01 (commented):
Added some responses to comments.

rfcs/0008-gateway-extension-communication-standard.md
AndrewKhoma: can't name it a system, I've checked the open PR's and found that the last one is 0009

rfcs/0008-gateway-extension-communication-standard.md
AndrewKhoma: @lichoil we need your help here, how safe is it to reference mongodb spec?

rfcs/0008-gateway-extension-communication-standard.md
AndrewKhoma: yes, so essentially type 0 section contains all args, except the one that allowed in Type 1 depending on request type

rfcs/0008-gateway-extension-communication-standard.md
AndrewKhoma: agree

rfcs/0008-gateway-extension-communication-standard.md
AndrewKhoma: it's inside the request_info, that is part of request_context

seoc01: I've updated rfc to be more generic with the example

PR #443, opened 2026-02-05
Instrumentation w/ Open Telemetry

udsmicrosoft:
Add OpenTelemetry instrumentation to DocumentDB gateway

Adds full OpenTelemetry support to the gateway, exporting traces, metrics, and logs via OTLP to any compatible backend.

Telemetry modules (pg_documentdb_gw/src/telemetry/)
- config.rs — Unified config with JSON (SetupConfiguration.json) → env var → default priority chain
- tracing.rs — TracerProvider with parent-based sampling (default 10%), batch span export
- metrics.rs — MeterProvider with counter-based metrics following OTel DB semantic conventions (db.client.operations, db.client.operation.duration.total, request/response size)
- logging.rs — LoggerProvider with OTLP export and optional console output
- telemetry_manager.rs — Initializes/shuts down all providers; gracefully degrades if collector is unreachable
- context_propagation.rs — W3C traceparent extraction from MongoDB comment field and injection into SQL comments

Gateway integration
- main.rs — Telemetry init on startup, graceful shutdown on exit
- lib.rs — handle_message span with activity_id, operation, db, collection, user_agent attributes
- data_management.rs — #[instrument] on processor functions
- setup.rs — TelemetryOptions deserialization from JSON config
- requests/mod.rs — Trace context extraction from incoming requests

Documentation
- docs/v1/telemetry.md — Configuration reference, available metrics/traces, context propagation, performance considerations, and local testing setup


docs/v1/telemetry.md
```diff
@@ -0,0 +1,275 @@
+# OpenTelemetry for DocumentDB Gateway
```
mehulj94, line 1: Can we make this into a rfc? Following the guideline here: https://github.com/documentdb/documentdb/blob/main/rfcs/0003-rfc-process.md

docs/v1/telemetry.md
```diff
@@ -0,0 +1,275 @@
...
+
+### Environment Variables
+
+| Variable | Default | Description |
+|----------|---------|-------------|
+| `OTEL_EXPORTER_OTLP_ENDPOINT` | `http://localhost:4317` | OTLP endpoint (gRPC) |
+| `OTEL_SERVICE_NAME` | `documentdb_gateway` | Service name in traces/metrics |
+| `OTEL_TRACING_ENABLED` | `false` | Enable distributed tracing |
+| `OTEL_METRICS_ENABLED` | `true` | Enable metrics export |
+| `OTEL_LOGGING_ENABLED` | `true` | Enable log export via OTLP |
```
mehulj94, line 32: With this do we keep generating logs in gateway.log and have an option to send to collector?

docs/v1/telemetry.md
```diff
@@ -0,0 +1,275 @@
...
+
+### Distributed Trace Context
+
+MongoDB wire protocol doesn't support HTTP-style trace headers. To enable end-to-end tracing, the gateway supports W3C trace context propagation via the MongoDB `comment` field:
+
+**1. Client passes trace context in the comment:**
+```javascript
+db.collection.find(
+  { status: "active" },
+  { comment: '{"traceparent": "00-abc123def456-789abc-01"}' }
```
mehulj94, line 168: instead of passing it as a string can we pass it as bson documentdb, for example: 
```
db.collection.find(
    { status: "active" },
    { comment: { traceparent: "00-abc123-def456-01" } }  // BSON document, not string
  )
```

pg_documentdb_gw/src/requests/mod.rs
```diff
@@ -417,6 +424,9 @@ impl<'a> Request<'a> {
                     }
                 }
                 "$readPreference" => ReadPreference::parse(v.as_document())?,
+                "comment" => {
+                    comment = v.as_str();
```
mehulj94: i think we should parse it as a bson document

pg_documentdb_gw/documentdb_gateway_core/src/telemetry/metrics.rs
```diff
@@ -0,0 +1,469 @@
...
+
+impl Default for OtelTelemetryProvider {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+#[async_trait::async_trait]
+impl crate::telemetry::TelemetryProvider for OtelTelemetryProvider {
+    async fn emit_request_event(
```
mehulj94: Since TelemetryManager::init_telemetry already registers the meter provider globally via global::set_meter_provider(), could we record these metrics directly in the request handling path instead of going through emit_request_event?                                                                                                                 
                                                                                                                                                                              
The tracing side of this PR already does this — #[instrument] and span attribute recording are in library code (handle_message, process_find, etc.) and run for every equest regardless of which TelemetryProvider is passed. Metrics could follow the same pattern: call global::meter() in the request flow, and if no meter provider is registered, it's a no-op.

This would keep TelemetryProvider as a pure extension hook for downstream consumers. Currently, run_gateway takes a single Option<Box<dyn TelemetryProvider>>, so a downstream consumer who wants both these OTel metrics and their own custom telemetry would need to write a composite provider that wraps OtelTelemetryProvider and forwards all parameters. If metrics were recorded directly in the library, downstream consumers could just pass their own provider and still get OTel metrics for free — and they could add their own custom OTel meters via global::meter() in their provider, all flowing through the same pipeline.

pg_documentdb_gw/src/lib.rs
mehulj94, line 780: All the data needed for OTel metrics (header, request_context.payload, response, request_context.tracker, collection) is already in scope here. The db.client.* counters from OtelTelemetryProvider could be recorded directly at this point instead of going through the trait — same as how span attributes are already recorded directly in handle_message above. Check my comment on metrics.rs for the full rationale.

pg_documentdb_gw/documentdb_gateway/src/main.rs
```diff
@@ -121,7 +131,18 @@ async fn start_gateway(setup_configuration: DocumentDBSetupConfiguration) {
         tls_provider,
     );
 
-    run_gateway::<DocumentDBDataClient>(service_context, None, shutdown_token)
+    let telemetry: Option<Box<dyn TelemetryProvider>> = if telemetry_initialized {
```
mehulj94: We should have a way to disable telemetry completely. From the current implementation telemetry_initialized will be true even if we turn off all three components in config (tracing, metrics, config)

pg_documentdb_gw/documentdb_gateway/src/main.rs
```diff
@@ -36,13 +37,6 @@ fn main() {
     let setup_configuration =
         DocumentDBSetupConfiguration::new(&cfg_file).expect("Failed to load configuration.");
 
-    tracing_subscriber::registry()
-        .with(EnvFilter::try_from_default_env().unwrap_or_else(|_| EnvFilter::new("info")))
-        .with(tracing_subscriber::fmt::layer())
```
mehulj94, line 39: Can we keep the console logging on by default? We dont want gateway to not write any logs if the otel collector is not running.

pg_documentdb_gw/src/telemetry/logging.rs
```diff
@@ -0,0 +1,353 @@
...
+    error::{DocumentDBError, Result},
+    telemetry::config::{env_var, DEFAULT_EXPORT_TIMEOUT_MS, DEFAULT_OTLP_ENDPOINT},
+};
+
+// ============================================================================
+// Constants
+// ============================================================================
+
+const DEFAULT_LOGGING_ENABLED: bool = true;
+const DEFAULT_CONSOLE_ENABLED: bool = false;
```
mehulj94: Can we make this true? If i am running gateway locally without an otel collector, i will get no log output at all

pg_documentdb_gw/documentdb_gateway_core/src/requests/mod.rs
```diff
@@ -417,6 +424,9 @@ impl<'a> Request<'a> {
                     }
                 }
                 "$readPreference" => ReadPreference::parse(v.as_document())?,
+                "comment" => {
```
mehulj94, line 442: Lets only parse this if tracing is enabled

pg_documentdb_gw/documentdb_gateway_core/src/processor/data_management.rs
```diff
@@ -38,6 +40,7 @@ pub async fn process_delete(
         .await
 }
 
+#[instrument(skip_all)]
```
mehulj94, line 43: Can we disable span creation if tracing is disabled?

pg_documentdb_gw/src/telemetry/metrics.rs
```diff
@@ -0,0 +1,469 @@
...
+            |ns: i64| -> f64 { Duration::from_nanos(ns.max(0) as u64).as_secs_f64() };
+
+        let duration_ns =
+            request_tracker.get_interval_elapsed_time(RequestIntervalKind::HandleRequest);
+
+        // Build attributes based on success/failure
+        let base_attrs: Vec<KeyValue> = match &response {
+            Either::Left(_) => {
+                vec![
+                    db_system.clone(),
```
mehulj94: Can we remove multiple string allocations here?

pg_documentdb_gw/src/telemetry/metrics.rs
```diff
@@ -0,0 +1,469 @@
...
+            Either::Right((_, size)) => *size as u64,
+        };
+        self.response_size_total
+            .add(response_size_bytes, &base_attrs);
+
+        // Record PostgreSQL phase breakdown (duration totals)
+        let pg_begin_ns = request_tracker
+            .get_interval_elapsed_time(RequestIntervalKind::PostgresBeginTransaction);
+        if pg_begin_ns > 0 {
+            let mut attrs = base_attrs.clone();
```
mehulj94: Is it possible to record without cloning here and other cloning in next few lines?

docs/v1/telemetry.md
udsmicrosoft: Console logging (stdout) is now on by default (OTEL_LOGS_CONSOLE_ENABLED=true), so the gateway always produces logs locally. OTLP export to a collector is additive and controlled separately via OTEL_LOGGING_ENABLED. Both can run simultaneously.

tsharp (commented):
@udsmicrosoft & @xgerman - this generally looks good but are you going to number the rfc? Also - for the open questions, can those be tracked as follow up items?

mehulj94:
@udsmicrosoft can we address the comment in metric.rs? Currently the PR blocks emit_request_event to be used as an extension hook

rfcs/XXXX-gateway-opentelemetry.md
tsharp: We could probably pull this RFC out and track this in a separate PR

PR #444, opened 2026-02-05
[RustGW] Add max_incoming_connections config for connection limiting

### LINKED ISSUE #432: [Feature Request] Add configurable maximum incoming connection limit to gateway
**Purpose of the feature.**

The gateway component accepts unlimited Mongo client connections. The accept loop spawns a Tokio task for every connection without any limit check, which can cause resource exhaustion (file descriptors, memory pressure) as well unwanted performance issues (unresponsive gateway, denial of service, connection storm issues). 

This feature does not have anything to do with backend PG query pool. Based on preliminary testing, the gateway can handle a large amount of TCP connections. Despite this, it would be good to have finer grain control on this limit.

This implementation uses a Tokio Semaphore to enforce connection limits at the TCP level. The approach adds a MaxIncomingConnections configuration option that controls how many concurrent connections the gateway accepts. 

Spec:

1) A semaphore is initialized with the maximum allowed connections at startup
2) When a new connection arrives, the server attempts to acquire a permit without blocking
3) If no permits are available, the TCP stream is dropped immediately (kernel sends RST to client)
4) The permit is held inside the spawned task and automatically released when the task exits
5) Permits are released even if the task panics, because Rust's Drop trait runs during stack unwinding

This implementation emulates MongoDB's behavior for consistency. When MongoDB exceeds connection limits, it logs:

```
{"t":{"$date":"2026-02-02T21:25:10.516+00:00"},"s":"I","c":"NETWORK","id":22942,"ctx":"listener","msg":"Connection refused because there are too many open connections","attr":{"remote":"240.10.0.1:38395","connectionCount":26214}}
```

1.) Atomic Counter with Scopeguard

Use an atomic counter to track active connections with a scopeguard ensuring cleanup. This is a lower-level implementation of the same concept as the semaphore approach.

Pros:

- Lower memory overhead (8 bytes total for the counter vs ~32 bytes per connection)
- Slightly less overhead per connection (no Arc clone)

Cons:

- Requires careful handling of memory ordering semantics (Acquire/Release/SeqCst)
- Manual panic safety implementation needed

2.) Send MongoDB Protocol Error

Accept connections and send TooManyLogicalSessions or similar BSON error before closing.

Pros:

- More explicit error message to clients

Cons:

- Wastes resources (Accept TCP, TLS, Auth, BSON serialization) on rejected connections
- Stops client retry mechanism

3.) Reverse Proxy / Load Balancer

Use external infrastructure (nginx, HAProxy, ALB/NLB, Envoy) to enforce connection limits.

Pros:

- Offloads connection management from application layer. 
- Can implement a rate-limit like solution.

Cons:

- Implementation is just for finer grain control on the server side
- Gateway already handles a lot of connections (~240k). 
- Additional infrastructure complexity and operational overhead

**Additional context**

Future Work:

- Telemetry enhancements (to be added in a future issue):
- Emit connection rejection events via TelemetryProvider
- Track metrics: total rejections, rejection rate, connection count over time

akshayrao96:
Closes #432

## Summary 

Without a connection limit, the gateway accepts unbounded TCP connections, which can exhaust memory and file descriptors on the host and overwhelm the Postgres backend during traffic spikes or misbehaving clients.               

Operators who use DocumentDB can now set a hard cap on concurrent connections, giving self-hosted and on-premise deployments predictable resource boundaries and protection against connection storms. This is a basic safeguard that production database deployments expect out of the box.   

## Usage

 Add `MaxIncomingConnections` to your gateway's `SetupConfiguration.json`:                                            
 
 ```
 {
    "MaxIncomingConnections": 1000 
}
```

## Latest Commit

Integration Test : `connection_limit_tests.rs`  It's a self-contained integration test that verifies the connection limit end-to-end. The test starts a gateway with max_incoming_connections: 1000 (configurable) and runs four phases:   
                                
1. Create Connections: open 500 clients (1000 connections), confirm each connects via ping. 
2. Reject Connections : attempt 50 clients more, assert each fails with ServerSelection timeout.
3. Recover : drop 50 clients, retry 1 client. Assert client succeeds. 
4. Reconnect : fill the rest with 49 clients back, confirm each succeeds.

## Changes

Add connection limiting to the gateway using a tokio Semaphore. When the configured limit is reached, new connections are refused with a warning log.

- Add `max_incoming_connections` config option to `SetupConfiguration` trait
- Add semaphore-based connection limiting in `run_gateway()`
- Default to `Semaphore::MAX_PERMITS` when not configured. This is `usize::MAX >> 3.`

## Testing
- Did manual testing. Latest commit has an integration test file

### CODE REVIEW

File: pg_documentdb_gw/src/context/service.rs [hunk 1]
```diff
@@ -23,6 +24,7 @@ pub struct ServiceContextInner {
     pub transaction_store: TransactionStore,
     pub query_catalog: QueryCatalog,
     pub tls_provider: TlsProvider,
+    pub connection_semaphore: Arc<Semaphore>,
```
AndrewKhoma: since this is purely a semaphore for the tcp connections, why do we add it here? can it be created in the place where we call it?

pg_documentdb_gw/src/lib.rs
```diff
@@ -190,12 +191,35 @@ where
...
+            result = tcp_listener.accept() => {
+                let (tcp_stream, addr) = match result {
+                    Ok(s) => s,
+                    Err(e) => {
+                        tracing::error!("Failed to accept connection: {e:?}");
+                        continue;
+                    }
+                };
+
+                let permit = match service_context.connection_semaphore().clone().try_acquire_owned() {
```
AndrewKhoma: seems like a logic is reversed here. maybe we should check the permit first and only then unwrap the tcp stream?

pg_documentdb_gw/src/lib.rs
```diff
@@ -190,12 +191,35 @@ where
...
+                    }
+                };
+
                 let conn_service_context = service_context.clone();
                 let conn_telemetry = telemetry.clone();
                 tokio::spawn(async move {
+                    let _permit = permit;
                     let conn_res = handle_connection::<T>(
-                        stream_and_address,
+                        Ok((tcp_stream, addr)),
```
AndrewKhoma: since you've already know that the tcp_stream and address are accepted, I would change the type here to just a tuple

pg_documentdb_gw/documentdb_gateway_core/src/lib.rs
```diff
@@ -190,12 +191,35 @@ where
...
+                        tracing::error!("Failed to accept connection: {e:?}");
+                        continue;
+                    }
+                };
+
+                let permit = match service_context.connection_semaphore().clone().try_acquire_owned() {
+                    Ok(p) => p,
+                    Err(_) => {
+                        let connection_count = service_context.setup_configuration().max_incoming_connections() - service_context.connection_semaphore().available_permits();
+                        drop(tcp_stream);
```
AndrewKhoma: but we're not informing client about an issue, I thought in issue we've discussed this and agreed on 
`{"t":{"$date":"2026-02-02T21:25:10.516+00:00"},"s":"I",  "c":"NETWORK",  "id":22942,   "ctx":"listener","msg":"Connection refused because there are too many open connections","attr":{"remote":"240.10.0.1:38395","connectionCount":26214}}` to be sent back to client

AndrewKhoma (changes requested):

pg_documentdb_gw/src/context/service.rs
akshayrao96: It could be created in the `run_gateway` function itself since it's only called there.

My thinking was `ServiceContext` is the place for shared state and that it would be easier to use from a maintainability perspective (if we add metrics or require it in different places or refactor code, since we have access to it if it's in that struct).

pg_documentdb_gw/src/lib.rs
akshayrao96: Good point. We could check the permit before unwrapping, I think it would work. The thing is I unwrap tcp stream first to also check if there even is a valid tcp stream to begin with. 

Before my diff, we don't handle the returned enum in `stream_and_address`, until its handled in the async `handle_connection` function here
` let (tcp_stream, peer_address) = stream_and_address?;`

My guess is if we have an err, then we would not want to do any permit logic anyway. So the way I did it was if we have a valid TCP stream, then check permit (try_acquire() is immediate, `ok()` if we have permits, `err()` if we don't in which we case we gracefully drop the tcp stream).

Let me know if I misinterpreted your comment.

pg_documentdb_gw/src/lib.rs
akshayrao96: This is in relation to the other comment. The `handle_connection` requires a result enum, so since I unwrapped the stream_and_address to a tuple, I need to wrap it back into a result enum.

The `stream_and_address` we have previously is actually of type Result. Alternatively since I unwrapped the tcp stream, checked that its valid, and obtained a permit for processing, I could change the method signature to have a tuple parameter instead. It would make this part redundant which I can remove then:

` let (tcp_stream, peer_address) = stream_and_address?;`

pg_documentdb_gw/documentdb_gateway_core/src/lib.rs
akshayrao96: It's just a server log (same as mongo). In mongo, the client does not get anything useful back, the mongo client just retries.

To actually send a response back to a client requires BSON construction, along with auth validation, which in my opinion is unnecessary computation work (unless we want to deviate from how mongo does it).

Also if we do send a response back to the client, it eliminates the retry logic that clients have which we may want to keep.

pg_documentdb_gw/src/context/service.rs
AndrewKhoma: ServiceContext is indeed a context for the whole gateway, but as you can see from above declarations in only stores the pointers to existing self-sufficient services, so I would add anything to serviceContext only when it's mature enough to have it's entry there

pg_documentdb_gw/src/lib.rs
AndrewKhoma: yes, we previously unwrapped there since we would send any error up to the run_gateway function and have tracing there:
```rust
if let Err(conn_err) = conn_res {
    tracing::error!("Failed to accept a TCP connection: {conn_err:?}.");
}
```

actually your intuition is right. let's check if we can accept the connection, then if permit is not allowing to do anything we need to place a message to the caller that we're closing the socket because of permit

pg_documentdb_gw/src/lib.rs
AndrewKhoma: please, change the method signature

pg_documentdb_gw/documentdb_gateway_core/src/lib.rs
AndrewKhoma: k, i see

can you give me the pointers where you saw it sends nothing back, please?
want to double-check

akshayrao96: Sure thing. I started a mongo and open source docdb server, both with tls. I included the changes in this commit for the open source docdb.

I set max connections to 12, and ran a python test script where I'm catching the exception and printing it when exceeding this limit.

Here both are matching from pymongo:

```
==================================================
  Target: mongo-tls | Port: 27018 | Max: 12
==================================================

  ✓ Client 1/12 connected [ss: 3]
  ✓ Client 2/12 connected [ss: 6]
  ✓ Client 3/12 connected [ss: 9]
  ✓ Client 4/12 connected [ss: 11]
  ✗ Client 5 failed: SSL handshake failed: localhost:27018: [SSL: UNEXPECTED_EOF_WHILE_READING] EOF occurred in violation of protocol (_ssl.c:1010) (configured timeouts: socketTimeoutMS: 5000.0ms, connectTimeoutMS: 5000.0ms), Timeout: 5.0s, Topology Description: <TopologyDescription id: 69a75eec87ffee225864adec, topology_type: Single, servers: [<ServerDescription ('localhost', 27018) server_type: Unknown, rtt: None, error=AutoReconnect('SSL handshake failed: localhost:27018: [SSL: UNEXPECTED_EOF_WHILE_READING] EOF occurred in violation of protocol (_ssl.c:1010) (configured timeouts: socketTimeoutMS: 5000.0ms, connectTimeoutMS: 5000.0ms)')>]>
```

```
==================================================
  Target: oss | Port: 27917 | Max: 12
==================================================

  ✓ Client 1/12 connected [ss: 2]
  ✓ Client 2/12 connected [ss: 4]
  ✓ Client 3/12 connected [ss: 6]
  ✓ Client 4/12 connected [ss: 8]
  ✓ Client 5/12 connected [ss: 10]
  ✗ Client 6 failed: SSL handshake failed: localhost:27917: [SSL: UNEXPECTED_EOF_WHILE_READING] EOF occurred in violation of protocol (_ssl.c:1010) (configured timeouts: socketTimeoutMS: 5000.0ms, connectTimeoutMS: 5000.0ms), Timeout: 5.0s, Topology Description: <TopologyDescription id: 69a75f963728b4f620a8a047, topology_type: Single, servers: [<ServerDescription ('localhost', 27917) server_type: Unknown, rtt: None, error=AutoReconnect('SSL handshake failed: localhost:27917: [SSL: UNEXPECTED_EOF_WHILE_READING] EOF occurred in violation of protocol (_ssl.c:1010) (configured timeouts: socketTimeoutMS: 5000.0ms, connectTimeoutMS: 5000.0ms)')>]>
```

The server doesn't send anything to the client, so no BSON sent to client. This is an error generated by pymongo client itself. 

As for server logs this is what I can see:

Mongo:
```
{"t":{"$date":"2026-03-03T22:34:22.992+00:00"},"s":"I",  "c":"NETWORK",  "id":22942,   "ctx":"listener","msg":"Connection refused because there are too many open connections","attr":{"remote":"240.10.0.1:60258","connectionCount":11}}
```

Open source docdb (the tracing format makes it hard to paste this in markdown)

```
[2m2026-03-03T22:30:15.018401Z[0m [33m WARN[0m [2mdocumentdb_gateway[0m[2m:[0m Connection refused because there are too many open connections [3mremote[0m[2m=[0m[::ffff:127.0.0.1]:54492 [3mconnectionCount[0m[2m=[0m10
```

AndrewKhoma: thanks for detailed explanation!

pg_documentdb_gw/documentdb_gateway_core/src/lib.rs
```diff
@@ -184,12 +192,35 @@ where
...
+                        tracing::error!("Failed to accept connection: {e:?}");
+                        continue;
+                    }
+                };
+
+                let permit = match connection_semaphore.clone().try_acquire_owned() {
+                    Ok(p) => p,
+                    Err(_) => {
+                        let connection_count = max_connections - connection_semaphore.available_permits();
+                        drop(tcp_stream);
```
AndrewKhoma: open-ended question: should we call shutdown (https://docs.rs/tokio/latest/tokio/io/trait.AsyncWriteExt.html#method.shutdown) before dropping?

akshayrao96: good shout! Let me read up more on this and account for the different scenarios this could help with

akshayrao96: I took a look at this and implemented the shutdown function to see the behavior. Server sends a FIN (finish) to the client to say "I'm done writing so gracefully close." Whereas drop just abruptly closes the connection, which triggers a RST (reset).

I don't think we should do it for the following:

1) MongoDB itself sends a RST in this scenario, so drivers already handle it as a retryable connection error.

2) During a connection storm, we want the rejection path to be as fast as possible. shutdown() is async so it would require spawning tasks with timeouts. It also initiates a 4-way handshake to gracefully close the connection.

Example scenario: if we have 10k permits and a client floods the system with connections (100k+), the server should drop as fast as possible to avoid the CPU overhead that shutdown() would incur per rejected connection.

AndrewKhoma: thank you for research and comparison!

AndrewKhoma (approved):

pg_documentdb_gw/documentdb_gateway_core/src/lib.rs
```diff
@@ -184,12 +192,35 @@ where
...
+            result = tcp_listener.accept() => {
+                let (tcp_stream, addr) = match result {
+                    Ok(s) => s,
+                    Err(e) => {
+                        tracing::error!("Failed to accept connection: {e:?}");
+                        continue;
+                    }
+                };
+
+                let permit = match connection_semaphore.clone().try_acquire_owned() {
```
AndrewKhoma: can we provide unit tests to validate this behavior?

akshayrao96: sure, let me look into the test infrastructure and add a test here. Will make a new pr here

AndrewKhoma: new iteration for this PR is good enough, let's not decouple tests from what we're testing

akshayrao96: Hey Andrew, apologies for the delay. Just did a rebase on my code and added tests for it. Also updated the summary. Would love to have it reviewed whenever you're free, thanks!

tsharp (approved):

pg_documentdb_gw/documentdb_gateway_core/src/lib.rs
```diff
@@ -38,6 +38,9 @@ use tokio_openssl::SslStream;
 use tokio_util::sync::CancellationToken;
 use uuid::Uuid;
 
+use std::sync::Arc;
+use tokio::sync::Semaphore;
```
AndrewKhoma: nit (feel free to dismiss since I'm working with @tsharp to automate this): but we are following the formatting of uses as StdExternalCrate, and group by crate, so `Semaphore` would fall under the `tokio::{` and `Arc` would be in `std::` clause

pg_documentdb_gw/documentdb_gateway_core/src/lib.rs
```diff
@@ -207,33 +217,76 @@ where
...
 {
+    let (tcp_stream, addr) = match stream_and_address {
+        Ok(s) => s,
+        Err(e) => {
+            tracing::error!("Failed to accept {protocol} connection: {e:?}");
+            return;
+        }
+    };
+
+    let permit = match connection_semaphore.clone().try_acquire_owned() {
```
AndrewKhoma: can you make it an explicit call of Arc::clone here, so we can justify why we borrow Arc

pg_documentdb_gw/documentdb_gateway_core/src/lib.rs
```diff
@@ -207,33 +217,76 @@ where
...
 fn spawn_unix_handler<T>(
     stream_result: std::io::Result<(UnixStream, UnixSocketAddr)>,
     service_context: ServiceContext,
     telemetry: Option<Box<dyn TelemetryProvider>>,
+    connection_semaphore: &Arc<Semaphore>,
+    max_connections: usize,
 ) where
     T: PgDataClient,
 {
+    let permit = match connection_semaphore.clone().try_acquire_owned() {
```
AndrewKhoma: Same as above: can we include Arc::clone here?

pg_documentdb_gw/documentdb_tests/tests/connection_limit_tests.rs
```diff
@@ -0,0 +1,188 @@
...
+                .allow_invalid_certificates(true)
+                .build(),
+        ))
+        .hosts(vec![ServerAddress::parse(format!("127.0.0.1:{PORT}"))?])
+        .max_pool_size(1)
+        .server_selection_timeout(Duration::from_secs(5))
+        .build();
+
+    Client::with_options(options)
+}
```
AndrewKhoma, line 94: but we have https://github.com/documentdb/documentdb/blob/main/pg_documentdb_gw/documentdb_tests/src/test_setup/clients.rs

can we reuse the `test_credentials` functions and also extend the existing ClientOptions builder functionality?

pg_documentdb_gw/documentdb_gateway_core/src/lib.rs
```diff
@@ -207,33 +217,76 @@ where
...
+            tracing::warn!(
+                connectionCount = connection_count,
+                "Unix socket connection refused because there are too many open connections"
+            );
+            return;
+        }
+    };
+
     tokio::spawn(async move {
+        let _permit = permit;
```
AndrewKhoma, line 289: can you document both here and above why we need to move permit to this scope, please?

I'm afraid that someone will remove it un-knowingly

safern (approved):

pg_documentdb_gw/documentdb_tests/tests/connection_limit_tests.rs
akshayrao96: yes that's a good idea. I've done so in the latest git commit on this PR

akshayrao96:
Thanks for the review! Went through all the comments:
  
  - Pulled from main so it's on latest changes.
  - Moved Arc/Semaphore into the existing std:: / tokio:: import groups (StdExternalCrate style)
  - Made the Arc::clone(...) explicit in both the TCP and Unix handlers
  - Shared client setup. Future callers can call this function and add more fields the client needs.
  - Testing. Ran the CI/CD locally.  all green, 358 passed / 0 failed, including the connection limit test. fmt/clippy
  (-D warnings) clean:
  
  ```
        Finished `test` profile [unoptimized + debuginfo] target(s) in 38.93s
       Running unittests src/main.rs (target/debug/deps/documentdb_gateway-e814d8056db50b29)
  test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
       Running unittests src/lib.rs (target/debug/deps/documentdb_gateway_core-cec9bbc8457d6cdd)
  test result: ok. 238 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.97s
       Running tests/command_tests.rs (target/debug/deps/command_tests-b42fd449e1111246)
  test result: ok. 25 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 5.24s
       Running tests/compact_tests.rs (target/debug/deps/compact_tests-7dc48ac4168f9222)
  test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.88s
       Running tests/connection_limit_tests.rs (target/debug/deps/connection_limit_tests-60e60311ea41aeaa)
  test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 312.18s
       Running tests/connection_pool_tests.rs (...)
  test result: ok. 7 passed; 0 failed; ...
  ... (one Running/result pair per suite) ...
       Running tests/user_crud_tests.rs (target/debug/deps/user_crud_tests-5a0f58499cc701ab)
  test result: ok. 16 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 2.21s
  ```
  
  **Edit: new commit ([8f098f6](https://github.com/documentdb/documentdb/pull/444/commits/8f098f6414e77269ee103a2165acac522860add5) ) reduced a timeout option from 5 to 0.5 seconds to make test run faster (takes 80 seconds). Otherwise it took 312 seconds as seen above.

AndrewKhoma (approved):

Issue #442, opened 2026-02-04
[Feature Request] Add support for more currentOp parameters


pamulart:
**Purpose of the feature.**
Add support for additional MongoDB `currentOp` parameters, which are currently supported in MongoDB but missing from our implementation: [](https://www.mongodb.com/docs/manual/reference/command/currentOp/)<https://www.mongodb.com/docs/manual/reference/command/currentOp/>

Currently Supported: command, active, ns, op, opid, secs_running\
Missing Parameters: lsid, transaction (txnNumber, autocommit, readConcern), cursor (cursorId, originatingCommand), client metadata, and microsecond-precision timing

These parameters are essential for debugging production issues and correlating operations with application logs.

**Describe the solution you'd like**
These parameters are not currently returned by the `currentOp` command, so we would like to implement infrastructure to support them.

Approach: Build shared memory array infrastructure in pg_documentdb extension to store operation metadata, enabling currentOp to return MongoDB-compatible fields.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #447, opened 2026-02-06
RFC-0011 Operation Metadata Tracking for currentOp Support

pamulart:
This RFC is in relation to https://github.com/documentdb/documentdb/issues/442

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -7,11 +7,11 @@
 | Abinav Rameesh | Microsoft |
 | Vinod Sridharan | Microsoft |
 | Siddhesh Vethe | Microsoft |
-| Zhe Li | Microsoft |
-| Rashim Gupta | Amazon |
-| Norman Townsend | Amazon |
+| German Eichberger | Microsoft |
 | Viktor Tarassov | Amazon |
+| Nitin Ahuja | Amazon |
 | Maruthi Devarenti | Amazon |
+| Vin Yu | Amazon |
 | Patrick McLain | AB InBev |
 | Hari Krishna | Rippling |
 | Hari Krishna Sunder | YugabyteDB |


rfcs/0011-Operation-Metadata-Tracking-for-currentOp-Support.md
```diff
@@ -0,0 +1,175 @@
...
+---
+
+## Approach
+### What is your proposed solution?
+The proposed solution is to use User Defined Functions (UDFs) in each path of an operation to set, update, and remove metadata from our currentOp data structure. The currentOp metadata will use a shared memory array to store operation metadata in the extension for easy access when clients query different currentOp parameters with minimal performance overhead.
+
+We will create `OpMetadataArray[MaxBackends]` in shared memory, where:
+
+* Each PostgreSQL backend has exactly one dedicated array slot
+* Backends write metadata to their own slot (`array[MyBackendId]`)
```
visridha: How does this get managed for stuff on a request basis?
I presume the detailed design section will cover such stuff.

pamulart: Yes the detailed design will go indepth on this.

But each backend has its unique id, and when the request arrives, it will be assigned to a backend connection. then the command starting will register the metadata, and when the command is finished it will clear the array.

I am still brainstorming the clearing of the metadata and where it should happen but this is the current concept.

Currently working on a poc for the insert command to determine performance overhead with this shared memory approach as well.

rfcs/0011-Operation-Metadata-Tracking-for-currentOp-Support.md
```diff
@@ -0,0 +1,300 @@
...
+### What are the key benefits and tradeoffs?
+
+**Benefits:**
+* This implementation allows us to add more parameters in the future to achieve MongoDB parity
+* Enables users to utilize parameters they expect from MongoDB
+* Provides a scalable foundation for extending monitoring capabilities
+
+**Tradeoffs:**
+* The gateway must not change the original command request from the client—any query optimization or transformation must happen in the extension layer, not the gateway
+* If future gateway transformations are needed, they must be carefully coordinated with currentOp metadata collection
```
AndrewKhoma: no, we should allow any gateway transformation, sine the gw is stateless

rfcs/0011-Operation-Metadata-Tracking-for-currentOp-Support.md
```diff
@@ -0,0 +1,300 @@
...
+```c
+typedef struct OpMetadata
+{
+	uint32_t commandLength;                   // Length of BSON data stored
+	char commandData[MAX_OP_COMMAND_LENGTH];  // Raw command BSON (1KB)
+	char lsid[MAX_LSID_LENGTH];              // Logical session ID (128 bytes)
+} OpMetadata;
+```
+
+__Global Shared Memory Array:__
```
AndrewKhoma: should it be an array or rather some hash-map?

rfcs/0011-Operation-Metadata-Tracking-for-currentOp-Support.md
```diff
@@ -0,0 +1,300 @@
...
+
+- No explicit cleanup required
+- Slot overwritten on next request
+- Backend termination handled by PostgreSQL
+
+
+
+### API Changes
+
+The currentOp command will have a number of new parameters added. For now only the lsid has been implemented to furhter performance impact testing.
```
AndrewKhoma: can you include which new params will be added?

rfcs/0011-Operation-Metadata-Tracking-for-currentOp-Support.md
```diff
@@ -0,0 +1,300 @@
...
+
+The testing will determine the performance impact of the new overhead involved with storing and retrieving metadata in the shared memory. This will include unit tests for the metadata storage and retrieval functions, integration tests to verify currentOp command compatibility with MongoDB behavior, and performance tests to measure the overhead under various load conditions. Additionally, compatibility tests will ensure that the new fields match MongoDB's field semantics and format requirements.
+
+### Documentation Updates
+
+Documentation that will need to be updated includes the addition of new currentOp parameter support and usage examples.
+
+---
+
+<!--## Implementation Tracking
```
AndrewKhoma: should we remove <!--?

rfcs/0011-Operation-Metadata-Tracking-for-currentOp-Support.md
pamulart: I think an array is sufficient because this data structure mimics the postgres' [backendstatusarray](https://doxygen.postgresql.org/backend__status_8h.html#a9f3cede341c58ad911bb12512c54ecb2) where the index will the the ith backend connection similar to this array.

rfcs/0011-Operation-Metadata-Tracking-for-currentOp-Support.md
pamulart: I see, rather we should avoid gateway implantation for the currentOp, having it in pg_documentdb is better for the aggregation stage.

PR #448, opened 2026-02-09
Add replica set fields support and isdbgrid configuration

### LINKED ISSUE #445: [Feature Request] Replica Set Support for DocumentDB
#### Problem Statement: 

DocumentDB currently lacks complete replica set support, which is a fundamental MongoDB feature that enables:

1. **High Availability** – Clearly identifying which node is the primary (writer) and which nodes are secondaries (readers)
2. **Read Scaling** – Allowing drivers to distribute read operations across secondary nodes using read preference settings
3. **Automatic Failover** – Enabling drivers to detect primary changes and reconnect automatically
4. **Standard Connection Strings** – Supporting replica set connection strings such as:

   ```
   mongodb://host1:27917,host2:27918/?replicaSet=rs0
   ```

When a MongoDB driver connects to a deployment, it executes the hello (or isMaster) command to determine cluster topology. For a replica set, this response must include:

- Replica set name (setName)

- List of member hosts (hosts)

- Current primary (primary)

- Identity of the responding node (me)

- Node role (isWritablePrimary / secondary) etc...

Additionally, the current hello response includes:

```
"msg": "isdbgrid"
```
This field is used by MongoDB drivers to identify a sharded cluster (mongos). Replica sets should not advertise themselves as sharded clusters. Including this field in non-sharded deployments may cause driver misclassification.

### 1. Support Optional Replica Set Metadata in hello / isMaster

Enhance the `hello`/`isMaster` command to optionally include replica set fields when replica set information is available through an extension or integration layer.

**For a Primary Node:**

```
{
  "isWritablePrimary": true,
  "setName": "rs0",
  "hosts": ["host1:27917", "host2:27918"],
  "primary": "host1:27917",
  "me": "host1:27917",
  "setVersion": 1,
  // ... other fields related to Replica set...
  "ok": 1
}
```

**For a Secondary Node:**
```
{
  "isWritablePrimary": false,
  "secondary": true,
  "setName": "rs0",
  "hosts": ["host1:27917", "host2:27918"],
  "primary": "host1:27917",
  "me": "host2:27918",
  // ... other fields related to Replica set...
  "ok": 1
}

```

#### Backward Compatibility

If no replica set metadata is available (i.e., no extension or integration provides it), DocumentDB should continue operating in standalone mode with current behavior:

```
{
  "isWritablePrimary": true,
  "maxBsonObjectSize": 16777216,
  "maxMessageSizeBytes": 48000000,
  "localTime": 1768599176805,
  "ok": 1
}
```

### 2. Make `isdbgrid` Behavior Configurable

Since `"msg": "isdbgrid"` indicates a sharded cluster (mongos), it should not appear in replica set deployments. Introduce a configuration option to control this behavior.

#### Proposed Configuration

Add a configuration option such as:
```json
{
  "IsMongoSharded": false
}
```

#### Behavior Based on Configuration

| `IsMongoSharded` Value | `hello` Response | `isdbgrid` Command |
|------------------------|------------------|-------------------|
| `false`| Does NOT include `"msg": "isdbgrid"` | Returns error: "no such cmd: isdbgrid" |
| `true`  (default)  | Includes `"msg": "isdbgrid"` | Returns `{ "isdbgrid": 1.0, "ok": 1 }` |

**When `IsMongoSharded: false` (Replica Set Mode):**
```json
{
  "isWritablePrimary": true,
  "setName": "rs0",
  // ... other fields ...
  // NO "msg": "isdbgrid"
  "ok": 1
}
```

This prevents drivers from incorrectly identifying replica set deployments as sharded clusters and ensures accurate deployment classification.

## Describe Alternatives Considered

1. Standalone-Only Behavior
Keeping DocumentDB strictly standalone simplifies implementation but limits compatibility with replica set–based deployments.

2. Hardcoded Replica Set Responses
Always returning replica set fields would reduce flexibility and may not reflect actual deployment topology.

The proposed approach provides flexibility while maintaining backward compatibility.

**Additional context**

While DocumentDB might not yet have full replica set capabilities, adding this extension point lays the groundwork. It ensures DocumentDB can evolve to support replica set topologies in the future or integrate with extensions that provide that functionality.

v-sathvik:
- Implement get_replica_set_fields function
- Add replica set fields SQL definitions
- Add isdbgrid configuration support in gateway
- Add isdbgrid integration tests

Fixes #445

### DIFF
--- a/pg_documentdb/sql/udfs/metadata/replica_set_fields--0.110-0.sql
+++ b/pg_documentdb/sql/udfs/metadata/replica_set_fields--0.110-0.sql
@@ -1,0 +1,10 @@
+/* Returns replica set fields for isMaster/hello response */
+
+CREATE OR REPLACE FUNCTION __API_SCHEMA_V2__.get_replica_set_fields()
+RETURNS __CORE_SCHEMA__.bson
+STABLE
+LANGUAGE c
+AS 'MODULE_PATHNAME', $function$get_replica_set_fields$function$;
+
+COMMENT ON FUNCTION __API_SCHEMA_V2__.get_replica_set_fields()
+	IS 'Returns replica set information as single BSON document.';
--- a/pg_documentdb/sql/udfs/metadata/replica_set_fields--0.110-0.sql
+++ b/pg_documentdb/sql/udfs/metadata/replica_set_fields--0.110-0.sql
@@ -1,10 +1,17 @@
-/* Returns replica set fields for isMaster/hello response */
-
-CREATE OR REPLACE FUNCTION __API_SCHEMA_V2__.get_replica_set_fields()
+/*
+ * Returns replica set fields for hello/isMaster responses.
+ *
+ * Returns NULL when no replica set hook is set, or a BSON document with replica set
+ * fields (hosts, setName, primary, me, isWritablePrimary, secondary, etc.) when a hook
+ * is configured. The gateway calls this periodically and merges the result into hello responses.
+ *
+ * Internal gateway use only.
+ */
+CREATE OR REPLACE FUNCTION documentdb_api_internal.get_replica_set_fields()
 RETURNS __CORE_SCHEMA__.bson
 STABLE
 LANGUAGE c
 AS 'MODULE_PATHNAME', $function$get_replica_set_fields$function$;
 
-COMMENT ON FUNCTION __API_SCHEMA_V2__.get_replica_set_fields()
+COMMENT ON FUNCTION documentdb_api_internal.get_replica_set_fields()
 	IS 'Returns replica set information as single BSON document.';
--- a/pg_documentdb/sql/udfs/metadata/replica_set_fields--0.110-0.sql
+++ b/pg_documentdb/sql/udfs/metadata/replica_set_fields--0.110-0.sql
@@ -5,7 +5,6 @@
  * fields (hosts, setName, primary, me, isWritablePrimary, secondary, etc.) when a hook
  * is configured. The gateway calls this periodically and merges the result into hello responses.
  *
- * Internal gateway use only.
  */
 CREATE OR REPLACE FUNCTION documentdb_api_internal.get_replica_set_fields()
 RETURNS __CORE_SCHEMA__.bson
--- a/pg_documentdb/sql/udfs/metadata/replica_set_fields--latest.sql
+++ b/pg_documentdb/sql/udfs/metadata/replica_set_fields--latest.sql
@@ -1,0 +1,10 @@
+/* Returns replica set fields for isMaster/hello response */
+
+CREATE OR REPLACE FUNCTION __API_SCHEMA_V2__.get_replica_set_fields()
+RETURNS __CORE_SCHEMA__.bson
+STABLE
+LANGUAGE c
+AS 'MODULE_PATHNAME', $function$get_replica_set_fields$function$;
+
+COMMENT ON FUNCTION __API_SCHEMA_V2__.get_replica_set_fields()
+	IS 'Returns replica set information as single BSON document.';
--- a/pg_documentdb/sql/udfs/metadata/replica_set_fields--latest.sql
+++ b/pg_documentdb/sql/udfs/metadata/replica_set_fields--latest.sql
@@ -1,10 +1,17 @@
-/* Returns replica set fields for isMaster/hello response */
-
-CREATE OR REPLACE FUNCTION __API_SCHEMA_V2__.get_replica_set_fields()
+/*
+ * Returns replica set fields for hello/isMaster responses.
+ *
+ * Returns NULL when no replica set hook is set, or a BSON document with replica set
+ * fields (hosts, setName, primary, me, isWritablePrimary, secondary, etc.) when a hook
+ * is configured. The gateway calls this periodically and merges the result into hello responses.
+ *
+ * Internal gateway use only.
+ */
+CREATE OR REPLACE FUNCTION documentdb_api_internal.get_replica_set_fields()
 RETURNS __CORE_SCHEMA__.bson
 STABLE
 LANGUAGE c
 AS 'MODULE_PATHNAME', $function$get_replica_set_fields$function$;
 
-COMMENT ON FUNCTION __API_SCHEMA_V2__.get_replica_set_fields()
+COMMENT ON FUNCTION documentdb_api_internal.get_replica_set_fields()
 	IS 'Returns replica set information as single BSON document.';
--- a/pg_documentdb/src/metadata/replica_set_fields.c
+++ b/pg_documentdb/src/metadata/replica_set_fields.c
@@ -1,0 +1,35 @@
+/*-------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation.  All rights reserved.
+ *
+ * src/metadata/replica_set_fields.c
+ *
+ * Implementation of replica set metadata functions.
+ *-------------------------------------------------------------------------
+ */
+#include <postgres.h>
+#include <fmgr.h>
+#include <nodes/parsenodes.h>
+
+#include "api_hooks.h"
+
+/* Function exports for SQL callable functions */
+PG_FUNCTION_INFO_V1(get_replica_set_fields);
+
+/*
+ * get_replica_set_fields
+ * 
+ * PostgreSQL callable function that returns replica set fields as a single BSON document.
+ * This function acts as a PostgreSQL interface that delegates to the hook system.
+ */
+Datum
+get_replica_set_fields(PG_FUNCTION_ARGS)
+{
+	Datum result = GetReplicaSetFields();
+	
+	if (result == (Datum) 0)
+	{
+		PG_RETURN_NULL();
+	}
+	
+	return result;
+}
--- a/pg_documentdb_gw/src/configuration/dynamic.rs
+++ b/pg_documentdb_gw/src/configuration/dynamic.rs
@@ -9,7 +9,7 @@
 use std::fmt::Debug;
 
 use async_trait::async_trait;
-use bson::RawBson;
+use bson::{RawBson, RawDocumentBuf};
 
 use crate::{configuration::Version, postgres};
 
@@ -27,6 +27,12 @@ pub trait DynamicConfiguration: Send + Sync + Debug {
     async fn enable_developer_explain(&self) -> bool;
     async fn max_connections(&self) -> usize;
     async fn allow_transaction_snapshot(&self) -> bool;
+
+    /// Returns replica set BSON for hello/isMaster responses.
+    /// Returns None when no replica set info is available.
+    async fn get_replica_set_bson(&self) -> Option<RawDocumentBuf> {
+        None
+    }
 
     // Needed to downcast to concrete type
     fn as_any(&self) -> &dyn std::any::Any;
--- a/pg_documentdb_gw/src/configuration/dynamic.rs
+++ b/pg_documentdb_gw/src/configuration/dynamic.rs
@@ -30,8 +30,13 @@ pub trait DynamicConfiguration: Send + Sync + Debug {
 
     /// Returns replica set BSON for hello/isMaster responses.
     /// Returns None when no replica set info is available.
-    async fn get_replica_set_bson(&self) -> Option<RawDocumentBuf> {
-        None
+    async fn get_replica_set_bson(&self) -> Option<RawDocumentBuf>;
+
+    /// Indicates whether to include "msg": "isdbgrid" in isMaster/hello responses
+    /// and whether the isdbgrid command returns success or error.
+    /// Default: true
+    async fn is_mongo_sharded(&self) -> bool {
+        self.get_bool("IsMongoSharded", true).await
     }
 
     // Needed to downcast to concrete type
--- a/pg_documentdb_gw/src/configuration/pg_configuration.rs
+++ b/pg_documentdb_gw/src/configuration/pg_configuration.rs
@@ -268,10 +268,6 @@ impl PgConfiguration {
         Ok(())
     }
 
-    pub async fn get_replica_set_bson(&self) -> Option<RawDocumentBuf> {
-        self.replica_set_bson.read().await.clone()
-    }
-
     pub fn start_config_watcher(
         dynamic_config_file: String,
         configuration: Arc<PgConfiguration>,
@@ -424,6 +420,10 @@ impl DynamicConfiguration for PgConfiguration {
 
     async fn allow_transaction_snapshot(&self) -> bool {
         self.get_bool("mongoAllowTransactionSnapshot", false).await
+    }
+
+    async fn get_replica_set_bson(&self) -> Option<RawDocumentBuf> {
+        self.replica_set_bson.read().await.clone()
     }
 
     fn as_any(&self) -> &dyn std::any::Any {
--- a/pg_documentdb_gw/src/configuration/pg_configuration.rs
+++ b/pg_documentdb_gw/src/configuration/pg_configuration.rs
@@ -16,7 +16,7 @@ use std::{
 };
 
 use async_trait::async_trait;
-use bson::{rawbson, RawBson};
+use bson::{rawbson, RawBson, RawDocumentBuf};
 use notify::{event::ModifyKind, Error, Event, EventKind, RecursiveMode, Watcher};
 use serde::Deserialize;
 use tokio::{
@@ -28,7 +28,7 @@ use tokio::{
 use crate::{
     configuration::{dynamic::POSTGRES_RECOVERY_KEY, DynamicConfiguration, SetupConfiguration},
     error::{DocumentDBError, Result},
-    postgres::{Connection, ConnectionPool, QueryCatalog},
+    postgres::{Connection, ConnectionPool, PgDocument, QueryCatalog},
     requests::request_tracker::RequestTracker,
 };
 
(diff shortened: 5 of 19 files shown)


pg_documentdb_gw/src/processor/constant.rs
```diff
@@ -48,11 +48,23 @@ pub fn process_get_cmd_line_opts() -> Result<Response> {
...
+        .is_mongo_sharded();
+
+    if is_enabled {
+        Ok(Response::Raw(RawResponse(rawdoc! {
+            "isdbgrid": 1.0,
+            "hostname": context.service_context.setup_configuration().node_host_name(),
+            "ok": OK_SUCCEEDED,
+        })))
+    } else {
+        Err(DocumentDBError::documentdb_error(
```
visridha: can we not just return isdbgrid: 0 and ok: 0?

pg_documentdb/src/metadata/replica_set_fields.c
```diff
@@ -0,0 +1,35 @@
...
+/*
+ * get_replica_set_fields
+ * 
+ * PostgreSQL callable function that returns replica set fields as a single BSON document.
+ * This function acts as a PostgreSQL interface that delegates to the hook system.
+ */
+Datum
+get_replica_set_fields(PG_FUNCTION_ARGS)
+{
+	Datum result = GetReplicaSetFields();
```
visridha, line 27: What's the point of this? and what's the point of having a function whose sole purpose is to return NULL?

pg_documentdb/sql/udfs/metadata/replica_set_fields--latest.sql
```diff
@@ -0,0 +1,10 @@
+/* Returns replica set fields for isMaster/hello response */
+
+CREATE OR REPLACE FUNCTION __API_SCHEMA_V2__.get_replica_set_fields()
```
visridha: What's the contract for this?

pg_documentdb_gw/src/configuration/pg_configuration.rs
```diff
@@ -216,11 +242,24 @@ impl PgConfiguration {
             }
         };
 
+        let new_replica_set_bson = match self.inner.load_replica_set_bson(conn).await {
```
visridha, line 257: what's the behavior here on null?

pg_documentdb/sql/udfs/metadata/replica_set_fields--0.110-0.sql
```diff
@@ -0,0 +1,10 @@
+/* Returns replica set fields for isMaster/hello response */
+
+CREATE OR REPLACE FUNCTION __API_SCHEMA_V2__.get_replica_set_fields()
```
visridha: for being a public API schema i don't get what the documented behavior of this function should be.

visridha (changes requested):

pg_documentdb_gw/src/configuration/pg_configuration.rs
```diff
@@ -114,6 +114,28 @@ impl PgConfigurationInner {
...
+                &request_tracker,
+            )
+            .await?;
+
+        if let Some(first_row) = replica_set_rows.first() {
+            let replica_set_doc: Option<PgDocument> = first_row.try_get(0)?;
+            if let Some(doc) = replica_set_doc {
+                return Ok(Some(doc.0.to_raw_document_buf()));
+            }
+        }
```
AndrewKhoma, line 145: should we move this loading logic to the load_configurations method and maybe we can store this info as part of the `PgConfigurationInner`?

pg_documentdb_gw/src/configuration/setup.rs
```diff
@@ -57,6 +57,13 @@ pub struct DocumentDBSetupConfiguration {
     // Unix socket file permissions (octal format string, e.g., "0660" for owner+group read/write)
     // If not specified, defaults to 0o660
     pub unix_socket_file_permissions: Option<String>,
+
+    #[serde(default = "default_is_mongo_sharded")]
+    pub is_mongo_sharded: Option<bool>,
```
AndrewKhoma: this can be bool, since you're providing the default loading function

pg_documentdb_gw/src/configuration/setup.rs
```diff
@@ -179,6 +186,10 @@ impl SetupConfiguration for DocumentDBSetupConfiguration {
         self.enforce_tls.unwrap_or(true)
     }
 
+    fn is_mongo_sharded(&self) -> bool {
+        self.is_mongo_sharded.unwrap_or(true)
```
AndrewKhoma: feels error-prone to me. why are we stating that mongo is sharded by default?

pg_documentdb_gw/src/processor/ismaster.rs
```diff
@@ -76,5 +83,16 @@ pub async fn process(
         );
     }
 
+    if let Some(pg_config) = dynamic_configuration
```
AndrewKhoma: should we add get_replica_set_bson inside of the trait?

pg_documentdb_gw/tests/isdbgrid_tests_disabled.rs
```diff
@@ -0,0 +1,53 @@
...
+    assert!(
+        result.is_err(),
+        "isdbgrid command should fail when disabled"
+    );
+
+    // Verify the error message contains the expected text
+    let error = result.err().unwrap();
+    let error_string = error.to_string();
+    assert!(
+        error_string.contains("no such cmd: isdbgrid") || error_string.contains("CommandNotFound"),
```
AndrewKhoma: but I don't see how do we block it anywhere, except actual processing, can we move it to be the part of the parsing?

pg_documentdb_gw/src/processor/constant.rs
AndrewKhoma: also it's weird to throw command not supported on the actual processing layer and not during parsing

AndrewKhoma (changes requested):

pg_documentdb_gw/src/processor/constant.rs
v-sathvik: Initially we thought the same, but looking at the MongoDB docs, it returns an error, not {ok: 0}:

https://www.mongodb.com/docs/manual/reference/command/isdbgrid/#:~:text=.%20The%20error-,document%20is%20similar%20to%20the%20following%3A,-%7B

pg_documentdb/src/metadata/replica_set_fields.c
v-sathvik: This function is a hook for replica set support. In the open-source document db, it returns NULL. In our extension, we implemented the hook to supply replica set fields. The gateway uses these fields in hello/isMaster responses.

pg_documentdb/sql/udfs/metadata/replica_set_fields--latest.sql
v-sathvik: The contract is, It returns a BSON document with replica set fields (hosts, setName, primary, me, etc.) when a hook is configured, or NULL when no hook is set . The gateway calls this periodically and merges the non-NULL result into hello/isMaster responses. Will move this to `documentdb_api_internal` since it's only used by the gateway, and add a proper contract comment in the SQL.

pg_documentdb_gw/src/configuration/pg_configuration.rs
v-sathvik: If the SQL function returns NULL, `load_replica_set_bson` returns `Ok(None)`. That `None` gets stored, and in ismaster.rs the `if let Some(pg_config)` block is skipped , so no replica set fields are added to the hello response.

pg_documentdb/sql/udfs/metadata/replica_set_fields--0.110-0.sql
v-sathvik: Agree, this shouldn't be in the public API schema. Will move it to `documentdb_api_internal` since it's only used by the gateway. so this returns replica set BSON when a hook is configured, NULL otherwise. Will document this properly.

pg_documentdb_gw/src/configuration/pg_configuration.rs
v-sathvik: I kept it separate because `load_configurations` returns `HashMap<String, String>` (GUC key-value settings), but this is `RawDocumentBuf` (binary BSON merged as-is into the hello response). Different types, different access patterns. Both get refreshed together in `reload_configuration_with_connection()` though.

pg_documentdb_gw/src/configuration/setup.rs
v-sathvik: You're right, will change to `bool` Thank you!

pg_documentdb_gw/src/configuration/setup.rs
v-sathvik: Default `true` is for backward compatibility,  before this change the gateway always included `"msg": "isdbgrid"` in hello and the isdbgrid command always worked. `true` keeps that existing behavior. `false` opts into replica set mode.

pg_documentdb_gw/src/processor/ismaster.rs
v-sathvik: Yeah, will add `get_replica_set_bson()` to the `DynamicConfiguration` trait with a default returning `None` and remove the downcast.

pg_documentdb_gw/src/processor/constant.rs
AndrewKhoma: 0 days without mongo acting weird...

pg_documentdb_gw/tests/isdbgrid_tests_disabled.rs
v-sathvik: You're right that the blocking happens in process_is_db_grid(). The parser uses Rust’s FromStr trait, which is stateless and only converts the command string to an enum, so it doesn’t have access to configuration.

The config (e.g., whether sharding is enabled) is only available in the processing layer via ConnectionContext. Moving the check to parsing would require passing SetupConfiguration through several parsing functions.

Keeping the check in processing keeps parsing config-independent. This should be fine as is. Let me know if you strongly prefer moving it earlier.

AndrewKhoma: SetupConfiguration is accessible through ServiceContext, which is a part of the ConnectionContext so we don't need to pass it anywhere

i don't have a very strong prefer of moving, but I like to have things organized logically together

pg_documentdb/sql/udfs/metadata/replica_set_fields--0.110-0.sql
```diff
@@ -0,0 +1,17 @@
+/*
+ * Returns replica set fields for hello/isMaster responses.
+ *
+ * Returns NULL when no replica set hook is set, or a BSON document with replica set
+ * fields (hosts, setName, primary, me, isWritablePrimary, secondary, etc.) when a hook
+ * is configured. The gateway calls this periodically and merges the result into hello responses.
+ *
+ * Internal gateway use only.
```
AndrewKhoma: can you expand on what do you mean by internal gateway use only?

pg_documentdb_gw/src/configuration/pg_configuration.rs
```diff
@@ -127,6 +149,7 @@ impl PgConfigurationInner {
 pub struct PgConfiguration {
     inner: PgConfigurationInner,
     values: RwLock<HashMap<String, String>>,
+    replica_set_bson: RwLock<Option<RawDocumentBuf>>,
```
AndrewKhoma: please, have it as a ArcSwap type, RwLock is an overkill for PgConfiguration where we read all of the time and reload it periodically

pg_documentdb_gw/src/lib.rs
```diff
@@ -640,6 +640,18 @@ where
...
     request_tracker.record_duration(RequestIntervalKind::FormatRequest, format_request_start);
 
+    if request.request_type() == &RequestType::IsDBGrid
+        && !connection_context
+            .service_context
+            .setup_configuration()
+            .is_mongo_sharded()
+    {
+        return Err(DocumentDBError::documentdb_error(
+            ErrorCode::CommandNotSupported,
```
AndrewKhoma, line 650: can we move this even a bit more earlier in `parse_cmd` so this error will be returned during the call to `protocol::reader::parse_request`

pg_documentdb_gw/src/configuration/dynamic.rs
```diff
@@ -28,6 +28,12 @@ pub trait DynamicConfiguration: Send + Sync + Debug {
     async fn max_connections(&self) -> usize;
     async fn allow_transaction_snapshot(&self) -> bool;
 
+    /// Returns replica set BSON for hello/isMaster responses.
+    /// Returns None when no replica set info is available.
+    async fn get_replica_set_bson(&self) -> Option<RawDocumentBuf> {
+        None
```
AndrewKhoma: why is it None by default? can we leave this trait method without default implementation?

pg_documentdb_gw/src/configuration/setup.rs
```diff
@@ -57,6 +57,13 @@ pub struct DocumentDBSetupConfiguration {
     // Unix socket file permissions (octal format string, e.g., "0660" for owner+group read/write)
     // If not specified, defaults to 0o660
     pub unix_socket_file_permissions: Option<String>,
+
+    #[serde(default = "default_is_mongo_sharded")]
+    pub is_mongo_sharded: bool,
```
AndrewKhoma: this still feels error-prone to me in terms of way who is setting it

since this is the setup config it's populated on a cluster lvl and this setting is for replica-set, e.g. per host, not per cluster. Can we move it to the hostconfig, please?

AndrewKhoma: @visridha please, correct me if I'm wrong, that's just my gut feeling

it shouldn't be defined on a setup config lvl since we populate the responses for ismaster per each host, not per the whole cluster

pg_documentdb_gw/tests/isdbgrid_tests_disabled.rs
AndrewKhoma, line 1: can we combine two tests together, please?

it feels like anti-pattern to me to have them separated.
i have a scratches of the diff in my backlog to make gw testing stateful, do we need to prioritize that change?

pg_documentdb_gw/tests/isdbgrid_tests_disabled.rs
```diff
@@ -0,0 +1,53 @@
...
+    // When disabled, isdbgrid command should fail with error (like mongod)
+    let result = db.run_command(doc! { "isdbgrid": 1 }).await;
+    assert!(
+        result.is_err(),
+        "isdbgrid command should fail when disabled"
+    );
+
+    // Verify the error message contains the expected text
+    let error = result.err().unwrap();
+    let error_string = error.to_string();
```
AndrewKhoma: let's also validate the return code is for command not supported here (see how we do it in `invalid_index_tests`)

AndrewKhoma (changes requested):

pg_documentdb/sql/udfs/metadata/replica_set_fields--0.110-0.sql
v-sathvik: When I said 'internal gateway use only,' I meant that external clients don't directly call this function. The gateway calls the SQL function at startup and periodically to refresh the cached replica set fields. The function uses the `get_replica_set_fields_hook` extension point (defined in api_hooks.c), other extensions can set this hook to provide their own replica set topology. I'll remove that wording to avoid confusion.

pg_documentdb_gw/src/configuration/pg_configuration.rs
v-sathvik: Ack.

pg_documentdb_gw/src/lib.rs
v-sathvik: I looked into this,  the parsing layer (parse_request → parse_msg → parse_cmd) is fully stateless right now and doesn't have access to any configuration. To move this check into parse_cmd, we'd need to thread DynamicConfiguration through the entire parse chain, which means changing a few function signatures plus all their call sites.

I placed it here to follow the same pattern as [send_shutdown_responses](https://github.com/documentdb/documentdb/blob/9e023e016e62ba71c21af394cbd4e3abc2ef7022/pg_documentdb_gw/src/lib.rs#L636) right above, which also does a config-dependent check after parsing. Seemed like the right spot to keep things consistent. Let me know if you'd still prefer moving it though.

pg_documentdb_gw/tests/isdbgrid_tests_disabled.rs
v-sathvik: no, combining the tests should be sufficient for now. Thank you!

pg_documentdb_gw/src/configuration/pg_configuration.rs
```diff
@@ -423,7 +432,7 @@ impl DynamicConfiguration for PgConfiguration {
     }
 
     async fn get_replica_set_bson(&self) -> Option<RawDocumentBuf> {
```
AndrewKhoma, line 434: async can be dropped

pg_documentdb_gw/tests/isdbgrid_tests_disabled.rs
AndrewKhoma: can we combine these 2 tests, please?

pg_documentdb_gw/tests/common/mod.rs
```diff
@@ -140,6 +140,28 @@ pub fn setup_configuration_with_unix_socket_custom(
...
+pub fn setup_configuration_with_dynamic_config(
+    dynamic_config_json: &str,
+) -> DocumentDBSetupConfiguration {
+    use std::fs;
+
+    // Create temp directory
+    let _ = fs::create_dir_all("/tmp");
+
+    // Create unique config file path
+    let config_path = format!("/tmp/test_dynamic_config_{}.json", std::process::id());
```
AndrewKhoma, line 154: can we have it with some uuid, instead of process id and also have some drop guard to cleanup after the test is done? i have an upcoming changes that is making each test having it's own gateway, so we can finally drop the `--test-threads=1`

AndrewKhoma (changes requested):

PR #452, opened 2026-02-10
[RFC-006] Phase 1: Add RBAC Metadata Infrastructure for Authorization

nitinahuja89:
Implements Phase 1 of RFC-006 by adding metadata infrastructure for Role-Based Access Control. This lays the groundwork for MongoDB-compatible authorization without modifying enforcement logic.

### DIFF
--- a/pg_documentdb/sql/rbac/builtin_roles--0.111-0.sql
+++ b/pg_documentdb/sql/rbac/builtin_roles--0.111-0.sql
@@ -1,0 +1,152 @@
+-- builtin_roles--0.111-0.sql
+-- Built-in Role Definitions
+-- Populates the roles table with standard DocumentDB built-in roles
+
+SET search_path TO documentdb_api_catalog, documentdb_core;
+
+-- Built-in Role: readWrite
+-- Provides read and write access to all collections in any database
+INSERT INTO documentdb_api_catalog.roles (document)
+SELECT documentdb_core.bson_json_to_bson('
+{
+  "_id": {"$oid": "6587f191e810c19729de8601"},
+  "role": "readWrite",
+  "is_builtin": true,
+  "description": "Provides read and write access to all collections",
+  "privileges": [
+    {
+      "resource": {"db": "", "collection": ""},
+      "actions": ["find", "insert", "update", "remove"]
+    }
+  ],
+  "roles": []
+}')
+WHERE NOT EXISTS (
+    SELECT 1 FROM documentdb_api_catalog.roles
+    WHERE document->>'role' = 'readWrite'
+);
+
+-- Built-in Role: read
+-- Provides read-only access to all collections in any database
+INSERT INTO documentdb_api_catalog.roles (document)
+SELECT documentdb_core.bson_json_to_bson('
+{
+  "_id": {"$oid": "6587f191e810c19729de8602"},
+  "role": "read",
+  "is_builtin": true,
+  "description": "Provides read-only access to all collections",
+  "privileges": [
+    {
+      "resource": {"db": "", "collection": ""},
+      "actions": ["find"]
+    }
+  ],
+  "roles": []
+}')
+WHERE NOT EXISTS (
+    SELECT 1 FROM documentdb_api_catalog.roles
+    WHERE document->>'role' = 'read'
+);
+
+-- Built-in Role: dbAdmin
+-- Provides database administration privileges
+INSERT INTO documentdb_api_catalog.roles (document)
+SELECT documentdb_core.bson_json_to_bson('
+{
+  "_id": {"$oid": "6587f191e810c19729de8603"},
+  "role": "dbAdmin",
+  "is_builtin": true,
+  "description": "Provides database administration privileges",
+  "privileges": [
+    {
+      "resource": {"db": "", "collection": ""},
+      "actions": [
+        "find", "insert", "update", "remove",
+        "createCollection", "dropCollection",
+        "createIndex", "dropIndex",
+        "collMod", "collStats", "dbStats",
+        "listCollections", "listIndexes"
+      ]
+    }
+  ],
+  "roles": []
+}')
+WHERE NOT EXISTS (
+    SELECT 1 FROM documentdb_api_catalog.roles
+    WHERE document->>'role' = 'dbAdmin'
+);
+
+-- Built-in Role: userAdmin
+-- Provides user and role administration privileges
+INSERT INTO documentdb_api_catalog.roles (document)
+SELECT documentdb_core.bson_json_to_bson('
+{
+  "_id": {"$oid": "6587f191e810c19729de8604"},
+  "role": "userAdmin",
+  "is_builtin": true,
+  "description": "Provides user and role administration privileges",
+  "privileges": [
+    {
+      "resource": {"db": "", "collection": ""},
+      "actions": [
+        "createUser", "dropUser", "updateUser",
+        "createRole", "dropRole", "updateRole",
+        "grantRole", "revokeRole",
+        "viewUser", "viewRole"
+      ]
+    }
+  ],
+  "roles": []
+}')
+WHERE NOT EXISTS (
+    SELECT 1 FROM documentdb_api_catalog.roles
+    WHERE document->>'role' = 'userAdmin'
+);
+
+-- Built-in Role: root
+-- Superuser role with all privileges (inherits from readWrite, dbAdmin, and userAdmin)
+INSERT INTO documentdb_api_catalog.roles (document)
+SELECT documentdb_core.bson_json_to_bson('
+{
+  "_id": {"$oid": "6587f191e810c19729de8605"},
+  "role": "root",
+  "is_builtin": true,
+  "description": "Superuser role with all privileges",
+  "privileges": [
+    {
+      "resource": {"db": "", "collection": ""},
+      "actions": [
+        "find", "insert", "update", "remove",
+        "createCollection", "dropCollection",
+        "createIndex", "dropIndex",
+        "collMod", "collStats", "dbStats",
+        "listCollections", "listIndexes", "listDatabases",
+        "createUser", "dropUser", "updateUser",
+        "createRole", "dropRole", "updateRole",
+        "grantRole", "revokeRole",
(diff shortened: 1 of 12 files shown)


pg_documentdb/sql/rbac/copy_role_grants--0.111-0.sql
```diff
@@ -0,0 +1,64 @@
...
+                WHEN r.rolname = 'documentdb_admin_role' THEN 'dbAdmin'
+                WHEN r.rolname = 'documentdb_user_admin_role' THEN 'userAdmin'
+                WHEN r.rolname = 'documentdb_root_role' THEN 'root'
+            END AS role_name,
+            'admin' AS database_name  -- Default to admin database for existing grants
+        FROM pg_auth_members am
+        JOIN pg_roles r ON r.oid = am.roleid
+        JOIN pg_roles u ON u.oid = am.member  -- Join to get username from member OID
+        WHERE r.rolname IN (
+            'documentdb_readwrite_role',
```
mehulj94, line 37: nit: maybe we can call this just "readwrite". Basically remove documentdb and role from the rolename

pg_documentdb/src/commands/users.c
```diff
@@ -1351,6 +1364,102 @@ IsCallingUserExternal()
 }
 
 
+/*
+ * PgRoleToMongodbRole maps PostgreSQL role names to DocumentDB role names
+ * for updating the RBAC metadata system.
+ */
+static const char *
+PgRoleToMongodbRole(const char *pgRole)
```
mehulj94, line 1372: if we store readWrite as the role name, then we dont need this lookup, will prevent string comparisons etc.

pg_documentdb/src/commands/users.c
```diff
@@ -1351,6 +1364,102 @@ IsCallingUserExternal()
...
+{
+	if (strcmp(pgRole, "documentdb_readwrite_role") == 0 ||
```
DeepakSPemmaraju, line 1374: Can we just compare with ApiReadWriteRole? It's value is supposed to be documentdb_readwrite_role anyways. Same question for the other mappings

pg_documentdb/src/commands/users.c
```diff
@@ -1351,6 +1364,102 @@ IsCallingUserExternal()
...
+	ExtensionExecuteQueryViaSPI(incrementVersion, readOnly, SPI_OK_SELECT, &isNull);
+}
+
+
+/*
+ * UpdateRbacMetadataForRevoke updates the RBAC metadata system by removing role grants.
+ * This should be called before dropping a user to remove all their role grants.
+ */
+static void
+UpdateRbacMetadataForRevoke(const char *username)
```
DeepakSPemmaraju, line 1444: Should we rename this UpdateRbacMetadataForDrop and have a separate method for Revoke? If I revoke a role from a user and don't drop the role entirely, I'm thinking we shouldn't delete the role from the roles table.

pg_documentdb_core/sql/rbac/001_metadata_tables.sql
```diff
@@ -0,0 +1,118 @@
...
+-- Create index on role name for efficient lookups
+-- Uses BSON arrow operator to extract 'role' field from document
+CREATE INDEX IF NOT EXISTS idx_roles_name 
+    ON documentdb_api_catalog.roles((document->>'role'));
+
+
+-- Grant access to user_roles table
+-- userAdmin and root can manage user role assignments
+GRANT SELECT, INSERT, UPDATE, DELETE ON documentdb_api_catalog.user_roles 
+    TO documentdb_user_admin_role, documentdb_root_role;
```
DeepakSPemmaraju: This roles doesn't exist in SQL yet, so we'll need to create it in this PR to do this grant

nitinahuja89: I have already removed this for now since they roles did not exist. We will do these grants once we add these roles. This should be included in the next commit I make.

pg_documentdb/sql/rbac/builtin_roles--0.111-0.sql
```diff
@@ -0,0 +1,150 @@
...
+-- Built-in Role Definitions
+-- Populates the roles table with standard DocumentDB built-in roles
+
+-- Built-in Role: readWrite
+-- Provides read and write access to all collections in any database
+INSERT INTO documentdb_api_catalog.roles (document)
+SELECT documentdb_core.bson_from_json('
+{
+  "_id": {"$oid": "6587f191e810c19729de8601"},
+  "role": "readWrite",
```
DeepakSPemmaraju, line 13: should we call this what its actually called in the system "documentdb_readwrite_role"? Same question for all the other roles below

pg_documentdb/sql/rbac/builtin_roles--0.111-0.sql
```diff
@@ -0,0 +1,150 @@
+-- 002_builtin_roles.sql
+-- Built-in Role Definitions
+-- Populates the roles table with standard DocumentDB built-in roles
+
+-- Built-in Role: readWrite
+-- Provides read and write access to all collections in any database
+INSERT INTO documentdb_api_catalog.roles (document)
+SELECT documentdb_core.bson_from_json('
+{
+  "_id": {"$oid": "6587f191e810c19729de8601"},
```
DeepakSPemmaraju, line 12: How are we getting this value? Same question for all the other roles below

pg_documentdb/sql/rbac/copy_role_grants--0.111-0.sql
```diff
@@ -0,0 +1,60 @@
+-- copy_role_grants--0.111-0.sql
```
seoc01, line 1: I don't know if we would want this to run during install, as it copies all users permissions into the mongo roles table, this seems like a choice an admin would want to make, and since we don't mirror any future changes from the pg_roles table, it feels like it could be confusing.

pg_documentdb/sql/rbac/metadata_tables--0.111-0.sql
```diff
@@ -0,0 +1,95 @@
+-- metadata_tables--0.111-0.sql
+-- RBAC Metadata Tables for RFC-006 Implementation
+-- Creates the core metadata infrastructure for RBAC
+
+SET search_path TO documentdb_api_catalog, documentdb_core;
+
+-- Table to track database-scoped role grants
+-- DocumentDB grants roles per-database: grantRolesToUser("alice", [{role: "readWrite", db: "sales"}])
+CREATE TABLE IF NOT EXISTS documentdb_api_catalog.user_roles (
+    username TEXT NOT NULL,           -- DocumentDB username (PostgreSQL role with LOGIN)
```
seoc01, line 10: Should we use OID to avoid issues with reusing names?

pg_documentdb/sql/rbac/metadata_tables--0.111-0.sql
```diff
@@ -0,0 +1,95 @@
...
+'Database scope for this role grant. User has this role only on this specific database.';
+
+-- Create index for database-level queries
+-- Note: username lookups use the PRIMARY KEY (username is leftmost column)
+CREATE INDEX IF NOT EXISTS idx_user_roles_database
+    ON documentdb_api_catalog.user_roles(database_name);
+
+-- Metadata versioning table for cache invalidation
+-- Stale cache can be detected by comparing cached version with current version
+CREATE TABLE IF NOT EXISTS documentdb_api_catalog.metadata_version (
```
seoc01, line 36: Did we review using an entry system table, to avoid going to storage every time we want to check the version?

pg_documentdb/sql/rbac/metadata_tables--0.111-0.sql
```diff
@@ -0,0 +1,95 @@
...
+
+-- Create index on role name for efficient lookups
+-- Uses BSON arrow operator to extract 'role' field from document
+CREATE INDEX IF NOT EXISTS idx_roles_name
+    ON documentdb_api_catalog.roles((document->>'role'));
+
+-- Create helper function to increment metadata version
+-- This function should be called after any RBAC metadata changes
+-- Note: BIGINT overflow is not handled as it would take 292,000+ years at 1000 increments/second
+CREATE OR REPLACE FUNCTION documentdb_api_catalog.increment_metadata_version()
```
seoc01: Should we use a trigger on the underlying tables to update the version number, rather than a helper function, so changes are reflected even if the tables are updates directly?

pg_documentdb/sql/rbac/builtin_roles--0.111-0.sql
```diff
@@ -0,0 +1,152 @@
...
+
+-- Built-in Role: read
+-- Provides read-only access to all collections in any database
+INSERT INTO documentdb_api_catalog.roles (document)
+SELECT documentdb_core.bson_json_to_bson('
+{
+  "_id": {"$oid": "6587f191e810c19729de8602"},
+  "role": "read",
+  "is_builtin": true,
+  "description": "Provides read-only access to all collections",
```
seoc01: I think we need to look at how we are modeling these, the "read" role needs to be scoped to a database, this looks like the readAnyDatabase, which can only be attached to the "admin" database.

pg_documentdb/sql/rbac/metadata_tables--0.111-0.sql
```diff
@@ -0,0 +1,95 @@
+-- metadata_tables--0.111-0.sql
+-- RBAC Metadata Tables for RFC-006 Implementation
+-- Creates the core metadata infrastructure for RBAC
+
+SET search_path TO documentdb_api_catalog, documentdb_core;
+
+-- Table to track database-scoped role grants
+-- DocumentDB grants roles per-database: grantRolesToUser("alice", [{role: "readWrite", db: "sales"}])
+CREATE TABLE IF NOT EXISTS documentdb_api_catalog.user_roles (
```
seoc01: All the other user tables store the info in bson format, I don't see a reason not do store this one this way,  just wondering if there is a specific reason?

seoc01 (commented):
Left a few comments around metadata structor

PR #454, opened 2026-02-12
Fix big endian build failures

lvoytek:
When building documentdb for s390x mainframe, which is a Big-Endian architecture, I ran into a few build issues. The first was a lack of definition for the local iterator variable in FillPgbsonElementUnsafe resulting in

```
src/io/pgbsonelement.c:301:40: error: ‘iterator’ undeclared (first use in this function); did you mean ‘inetor’?
  301 |         if (!bson_iter_init_from_data(&iterator,
      |                                        ^~~~~~~~
      |                                        inetor
```

The second was for the definition of WKB_BYTE_ORDER which had an extra `=`

```
include/geospatial/bson_geospatial_private.h:36:25: error: expected expression before ‘=’ token
   36 | # define WKB_BYTE_ORDER = (unsigned char) 0
      |                         ^
```

With both issues fixed s390x builds succeed again.


AndrewKhoma (approved):
thanks for building it!

PR #468, opened 2026-02-24
Implement rename_collection function overloading for full BSON support

v-sathvik:
- Add BSON signature for rename_collection in Extension
- Update Gateway to pass full BSON for rename_collection
- Add rename_collection BSON regression tests

#### Summary: 
Currently the Gateway extracted 4 scalar fields (source_db, source_collection, target_collection, drop_target) from the client's OP_MSG BSON and passed them individually to the Extension. This discarded session info (lsid), writeConcern, and other fields.

This change adds a new BSON-accepting signature `rename_collection(bson)` so the Gateway passes the complete client BSON to the Extension. The Extension parses the BSON internally to extract the needed fields. The legacy text signature `rename_collection(text, text, text, bool)` is preserved for backward compatibility.

Related to #440


pg_documentdb/sql/udfs/schema_mgmt/rename_collection--0.110-0.sql
```diff
@@ -24,4 +24,23 @@ LANGUAGE c
...
+CREATE OR REPLACE FUNCTION __API_SCHEMA_V2__.rename_collection(
+    p_commandspec __CORE_SCHEMA_V2__.bson)
+RETURNS void
+LANGUAGE c
+VOLATILE PARALLEL UNSAFE
+AS 'MODULE_PATHNAME', $function$command_rename_collection_by_bson_spec$function$;
+COMMENT ON FUNCTION __API_SCHEMA_V2__.rename_collection(__CORE_SCHEMA_V2__.bson)
+    IS 'rename a collection';
+
+CREATE OR REPLACE FUNCTION documentdb_api_v2.rename_collection(
```
AndrewKhoma, line 39: this is question to @safern and @visridha, why do we need documentdb_api_v2 if we already declared the function with __API_SCHEMA_V2__?

pg_documentdb/src/commands/rename_collection.c
```diff
@@ -21,38 +21,108 @@
...
+typedef struct RenameCollectionArgs
+{
+	char *databaseName;
+	char *sourceCollectionName;
+	char *targetCollectionName;
+	bool dropTarget;
+} RenameCollectionArgs;
+
+static void ParseRenameCollectionSpec(pgbson *commandSpec, RenameCollectionArgs *args);
+static void ExecuteRenameCollection(char *databaseName, char *sourceCollectionName,
```
AndrewKhoma: can we share this function (e.g. put it somewhere in .h file and drop static here?)

pg_documentdb/src/test/regress/sql/rename_collection_bson_spec.sql
```diff
@@ -0,0 +1,48 @@
...
+-- verify old name gone, new name has data
+SELECT count(*) FROM documentdb_api.collection('rename_bson_db','source_coll');
+SELECT count(*) FROM documentdb_api.collection('rename_bson_db','target_coll');
+
+-- Test 2: rename to existing name without dropTarget → ERROR
+SELECT documentdb_api.create_collection('rename_bson_db','existing_coll');
+SELECT documentdb_api.rename_collection('{"renameCollection": "rename_bson_db.target_coll", "to": "rename_bson_db.existing_coll"}'::documentdb_core.bson);
+
+-- Test 3: rename with dropTarget=true → succeeds
+SELECT documentdb_api.rename_collection('{"renameCollection": "rename_bson_db.target_coll", "to": "rename_bson_db.existing_coll", "dropTarget": true}'::documentdb_core.bson);
```
AndrewKhoma, line 22: should we also change the other sql tests to use a new bson body? so we can retire the old rename_collection in 2 releases?

AndrewKhoma: i see at least one such in `collection_management.sql`

pg_documentdb_gw/src/processor/data_description.rs
```diff
@@ -169,14 +165,7 @@ pub async fn process_rename_collection(
     }
```
AndrewKhoma: should we move this logic to the backend too? so the gateway remains slim and don't do any payload parsing at all?

pg_documentdb/sql/udfs/schema_mgmt/rename_collection--0.110-0.sql
```diff
@@ -24,4 +24,23 @@ LANGUAGE c
...
+    p_commandspec __CORE_SCHEMA_V2__.bson)
+RETURNS void
```
visridha: Why not return the bson spec response (in case we need to update this in the future to return something more than ok: 1

pg_documentdb/src/commands/rename_collection.c
```diff
@@ -21,38 +21,108 @@
...
+			targetNamespace = pstrdup(element.bsonValue.value.v_utf8.str);
+		}
+		else if (strcmp(element.path, "dropTarget") == 0)
+		{
+			EnsureTopLevelFieldType("dropTarget", &specIter, BSON_TYPE_BOOL);
+			args->dropTarget = element.bsonValue.value.v_bool;
+		}
+		else if (!IsCommonSpecIgnoredField(element.path))
+		{
+			elog(DEBUG1, "Unrecognized command field: renameCollection.%s", element.path);
```
visridha: Any reason not to fail on unrecognized fields?

visridha: Also nit on style: prefer to use ereport than elog

pg_documentdb/src/commands/rename_collection.c
```diff
@@ -21,38 +21,108 @@
...
+	if (sourceNamespace == NULL || targetNamespace == NULL)
 	{
-		ereport(ERROR, (errmsg("collection target name cannot be NULL")));
+		ereport(ERROR, (errcode(ERRCODE_DOCUMENTDB_BADVALUE),
+						errmsg("renameCollection command must specify both renameCollection and to fields")));
+	}
+
+	/* Split "db.collection" into separate parts */
+	char *sourceDot = strchr(sourceNamespace, '.');
+	char *targetDot = strchr(targetNamespace, '.');
```
visridha: Do we need to validate sourceDB == targetDB?

pg_documentdb/src/commands/rename_collection.c
```diff
@@ -21,38 +21,108 @@
...
 	}
 
-	Datum database_datum = PG_GETARG_DATUM(0);
-	Datum collection_datum = PG_GETARG_DATUM(1);
-	Datum new_collection_datum = PG_GETARG_DATUM(2);
-	bool drop_target = PG_ARGISNULL(3) ? false : PG_GETARG_BOOL(3);
+	*sourceDot = '\0';
+	*targetDot = '\0';
+
+	args->databaseName = sourceNamespace;
```
visridha: sourceNamespace is the full thing with the dots right?

pg_documentdb/src/commands/rename_collection.c
```diff
@@ -21,38 +21,108 @@
...
-	if (PG_ARGISNULL(2))
+	if (sourceNamespace == NULL || targetNamespace == NULL)
 	{
-		ereport(ERROR, (errmsg("collection target name cannot be NULL")));
+		ereport(ERROR, (errcode(ERRCODE_DOCUMENTDB_BADVALUE),
+						errmsg("renameCollection command must specify both renameCollection and to fields")));
+	}
+
+	/* Split "db.collection" into separate parts */
+	char *sourceDot = strchr(sourceNamespace, '.');
```
visridha: nit: If you use StringView there's convenience APIs for FindSubstring/Suffix etc that deal with this stuff.

pg_documentdb/src/commands/rename_collection.c
```diff
@@ -21,38 +21,108 @@
...
+	{
+		ereport(ERROR, (errcode(ERRCODE_DOCUMENTDB_BADVALUE),
+						errmsg("Invalid namespace format, expected 'database.collection'")));
 	}
 
-	Datum database_datum = PG_GETARG_DATUM(0);
-	Datum collection_datum = PG_GETARG_DATUM(1);
-	Datum new_collection_datum = PG_GETARG_DATUM(2);
-	bool drop_target = PG_ARGISNULL(3) ? false : PG_GETARG_BOOL(3);
+	*sourceDot = '\0';
```
visridha: Why mutate the string?

pg_documentdb_gw/src/processor/data_description.rs
visridha: Move these validations to the backend and remove this logic from here?

pg_documentdb/src/commands/rename_collection.c
v-sathvik: Agreed.

pg_documentdb/src/test/regress/sql/rename_collection_bson_spec.sql
v-sathvik: We already have BSON specific tests in rename_collection_bson_spec.sql which covers all the cases. Regarding retiring the old text signature , we kept it intentionally because metadata/collection.c:RenameCollection() is a C-level wrapper that calls the 4-param text signature via SPI.  So the old signature cannot be retired yet without also updating that wrapper. Should we do that as part of this PR, or track it as a follow-up?

pg_documentdb_gw/src/processor/data_description.rs
v-sathvik: Agreed. Yes make sense.

pg_documentdb/sql/udfs/schema_mgmt/rename_collection--0.110-0.sql
v-sathvik: Agreed! We initially followed the same RETURNS void pattern as the existing signature, but returning bson makes sense  for future extensibility. Fixed, the BSON overload now returns bson with { "ok": 1.0 }.

pg_documentdb/src/commands/rename_collection.c
v-sathvik: I will address this in the next revision. After the fix, Unknown fields now fail with an UNKNOWNBSONFIELD error.

pg_documentdb/src/commands/rename_collection.c
v-sathvik: Yes,  added the validation right after the split. Without it, the target database name gets discarded and cross-database rename silently happens within the source database, returning { ok: 1 } with wrong behavior.

pg_documentdb/src/commands/rename_collection.c
v-sathvik: Yes, correct sourceNamespace starts as the full "db.collection" string. We overwrite the dot with \0 so it reads as just the db name. But this might be confusing , will replace this approach with StringView as suggested in the below comment.

pg_documentdb/src/commands/rename_collection.c
v-sathvik: Agreed, we didn't need to. Replaced with StringView which avoids the mutation entirely.

pg_documentdb/src/commands/rename_collection.c
v-sathvik: Agreed.

pg_documentdb_gw/src/processor/data_description.rs
v-sathvik: Ack.

pg_documentdb/src/test/regress/sql/rename_collection_bson_spec.sql
AndrewKhoma: let's do it in a follow-up and create a tracking issue/work item to retire old signature

AndrewKhoma (commented):
thanks for your changes! can you, please, rebase to the latest main and populate some tests in pg_documentdb_gw/documentdb_tests/ to ensure that the schema is correctly defined in gateway and add some sanity tests to see the error propagation from backend to gateway, thanks!

pg_documentdb_gw/documentdb_tests/src/commands/rename_collection.rs
```diff
@@ -0,0 +1,132 @@
...
+        .await?;
+    db.collection::<Document>("coll_b")
+        .insert_one(doc! { "_id": 2, "from": "b" })
+        .await?;
+
+    let result = db
+        .run_command(doc! {
+            "renameCollection": ns(db, "coll_a"),
+            "to": ns(db, "coll_b"),
+            "dropTarget": true,
```
AndrewKhoma, line 62: let's also validate that the original target was dropped

pg_documentdb/sql/udfs/schema_mgmt/rename_collection--0.110-0.sql
safern: This is for RBAC support that is in development. While that is in development we need to have both APIs because of the different GRANTs between the roles

pg_documentdb_gw/documentdb_gateway_core/src/processor/data_description.rs
```diff
@@ -112,69 +112,8 @@ pub async fn process_rename_collection(
...
     pg_data_client
-        .execute_rename_collection(
-            request_context,
-            source_db,
-            source_coll,
-            target_coll,
-            drop_target,
-            connection_context,
-        )
+        .execute_rename_collection(request_context, connection_context)
```
AndrewKhoma, line 116: i'm thinking about the cases if someone is running fresh gateway, but haven't updated their pg extension should we add some sort of version check saying that if the version is less than 110, then do the old style and if the version is equal or greater than 110 then follow a new code path

if that's too much to ask we can leave it as a follow-up, cause it's more of the infra task that is not implemented in this codebase. we can chat more about this over the Discord, lmk

pg_documentdb/include/commands/rename_collection.h
```diff
@@ -0,0 +1,31 @@
...
+{
+	char *databaseName;
+	char *sourceCollectionName;
+	char *targetCollectionName;
+	bool dropTarget;
+} RenameCollectionArgs;
+
+void ParseRenameCollectionSpec(pgbson *commandSpec, RenameCollectionArgs *args);
+void ExecuteRenameCollection(char *databaseName, char *sourceCollectionName,
+							 char *targetCollectionName, bool dropTarget);
```
safern, line 29: Do we need to export this on a header file? It seems like all these will just be used by rename_collection and a single file, so I don't think it needs to be exported.

pg_documentdb_gw/documentdb_gateway_core/src/processor/data_description.rs
v-sathvik: Yes, lets do this as a follow up,  we can discuss this over Discord.

pg_documentdb/include/commands/rename_collection.h
v-sathvik: Initially I had implemented it the same way,  keeping the functions static within rename_collection.c without a header. But  a previous comment on this PR (https://github.com/documentdb/documentdb/pull/468#discussion_r2849707523)  suggested sharing these functions by putting  them in a header file, so we implemented it that way. 

Can revert back to the original approach,  let me know which direction you'd prefer.

pg_documentdb_gw/documentdb_tests/src/commands/rename_collection.rs
```diff
@@ -0,0 +1,143 @@
...
+        })
+        .await?;
+
+    assert_eq!(result.get_f64("ok").unwrap(), 1.0);
+
+    let doc = db
+        .collection::<Document>("coll_b")
+        .find_one(doc! {})
+        .await?
+        .unwrap();
```
AndrewKhoma, line 72: nit: why do we need unwrap here?

AndrewKhoma (approved):

pg_documentdb_gw/documentdb_tests/src/commands/rename_collection.rs
v-sathvik: find_one returns `Option<Document>`, so after `.await?` we have an Option. We need to unwrap it to extract the Document before we can call doc.get_str("from") on the next line.

pg_documentdb/include/commands/rename_collection.h
safern: I would leave this to whatever you prefer to do it, either way works for us. Thanks for clarifying!

v-sathvik: Thank you! In that case, I will leave it as it is.

visridha (approved):

safern (approved):

PR #472, opened 2026-02-26
Implemented listDatabases command for gateway standardization - RFC:0010

drk-knght:
## Description
Related to https://github.com/documentdb/documentdb/pull/440
Implemented full BSON support for the listDatabases command by replacing inline SQL query construction in the gateway with a call to the database API - `SELECT documentdb_api.list_databases`

listDatabases command can take 4 different optional parameters- `filter, nameOnly, authorizedDatabases, comment`,  currently the existing API supports only the `filter` option so the SQL tests are added accordingly to check the correctness of the functionality.

### Mongo Test
```
[direct: mongos] test> use testdb1
switched to db testdb1
[direct: mongos] testdb1> db.users.insertOne({ name: "Alice", age: 30, city: "New York" })
{
  acknowledged: true,
  insertedId: ObjectId('6996601cde93b2b6a78ce5b0')
}
[direct: mongos] testdb1> db.users.insertOne({ name: "Bob", age: 25, city: "Boston" })
{
  acknowledged: true,
  insertedId: ObjectId('69966022de93b2b6a78ce5b1')
}
[direct: mongos] testdb1> db.products.insertOne({ name: "Laptop", price: 999 })
{
  acknowledged: true,
  insertedId: ObjectId('69966027de93b2b6a78ce5b2')
}
[direct: mongos] testdb1> use testdb2
switched to db testdb2
[direct: mongos] testdb2> db.orders.insertOne({ orderId: 1, amount: 150, status: "completed" })
{
  acknowledged: true,
  insertedId: ObjectId('69966037de93b2b6a78ce5b3')
}
[direct: mongos] testdb2> db.orders.insertOne({ orderId: 2, amount: 200, status: "pending" })
{
  acknowledged: true,
  insertedId: ObjectId('6996603dde93b2b6a78ce5b4')
}
[direct: mongos] testdb2> db.customers.insertOne({ customerId: 1, name: "Charlie" })
{
  acknowledged: true,
  insertedId: ObjectId('69966041de93b2b6a78ce5b5')
}

[direct: mongos] testdb2> db.adminCommand({ listDatabases: 1 })
{
  databases: [
    { name: 'testdb2', sizeOnDisk: 0, empty: false },
    { name: 'testdb1', sizeOnDisk: 0, empty: false }
  ],
  totalSize: 18736407,
  ok: 1
}

[direct: mongos] testdb2> db.adminCommand({ listDatabases: 1, nameOnly: false })
{
  databases: [
    { name: 'testdb2', sizeOnDisk: 0, empty: false },
    { name: 'testdb1', sizeOnDisk: 0, empty: false }
  ],
  totalSize: 18744599,
  ok: 1
}

[direct: mongos] testdb2> db.adminCommand({ listDatabases: 1, nameOnly: true })
{ databases: [ { name: 'testdb2' }, { name: 'testdb1' } ], ok: 1 }
```

### SQL Test
```
ok 233       - commands_db_stats                       13266 ms
ok 234       - command_list_databases                    241 ms
ok 235       - commands_create_view_tests                322 ms
```

### DIFF
--- a/pg_documentdb_gw/src/postgres/documentdb_data_client.rs
+++ b/pg_documentdb_gw/src/postgres/documentdb_data_client.rs
@@ -588,39 +588,22 @@ impl PgDataClient for DocumentDBDataClient {
     ) -> Result<Response> {
         let (request, request_info, request_tracker) = request_context.get_components();
         // TODO: Handle the case where !nameOnly - the legacy gateway simply returns 0s in the appropriate format
-        let filter = request.document().get_document("filter").ok();
-        let filter_string = filter.map_or("", |_| "WHERE document @@ $1");
-
         let list_db_query = connection_context
             .service_context
             .query_catalog()
-            .list_databases(filter_string);
-        let connection = self.pull_connection(connection_context).await?;
+            .list_databases();
 
-        let list_database_rows = match filter {
-            None => {
-                connection
-                    .query(
-                        &list_db_query,
-                        &[],
-                        &[],
-                        Timeout::transaction(request_info.max_time_ms),
-                        request_tracker,
-                    )
-                    .await?
-            }
-            Some(filter) => {
-                connection
-                    .query(
-                        &list_db_query,
-                        &[Type::BYTEA],
-                        &[&PgDocument(filter)],
-                        Timeout::transaction(request_info.max_time_ms),
-                        request_tracker,
-                    )
-                    .await?
-            }
-        };
+        let list_database_rows = self
+            .pull_connection(connection_context)
+            .await?
+            .query(
+                list_db_query,
+                &[Type::BYTEA],
+                &[&PgDocument(request.document())],
+                Timeout::transaction(request_info.max_time_ms),
+                request_tracker,
+            )
+            .await?;
 
         Ok(Response::Pg(PgResponse::new(list_database_rows)))
     }
--- a/pg_documentdb_gw/src/postgres/query_catalog.rs
+++ b/pg_documentdb_gw/src/postgres/query_catalog.rs
@@ -310,9 +310,8 @@ impl QueryCatalog {
         &self.update_bulk
     }
 
-    pub fn list_databases(&self, filter_string: &str) -> String {
-        self.list_databases
-            .replace("{filter_string}", filter_string)
+    pub fn list_databases(&self) -> &str {
+        &self.list_databases
     }
 
     pub fn list_collections(&self) -> &str {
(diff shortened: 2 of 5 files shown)


visridha:
Could you please also add/replicate tests under pg_documentdb for non distributed test coverage?

AndrewKhoma (commented):
can we add more tests to `pg_documentdb_gw/documentdb_tests/src/commands/list_databases.rs`?

AndrewKhoma:
overall LGTM

visridha:
Will wait for @AndrewKhoma's approval on this one.

pg_documentdb_gw/src/postgres/query_catalog.rs
```diff
@@ -511,15 +510,7 @@ pub fn create_query_catalog() -> QueryCatalog {
...
-            list_databases: "WITH r1 AS (SELECT DISTINCT database_name AS name
-                                FROM documentdb_api_catalog.collections),
-                             r2 AS (SELECT documentdb_core.row_get_bson(r1) AS document FROM r1),
-                             r3 AS (SELECT document FROM r2 {filter_string}),
-                             r4 AS (SELECT COALESCE(documentdb_api_catalog.bson_array_agg(r3.document, ''), '{ \"\": [] }') AS \"databases\",
-                                           1.0::float8                                                                        AS \"ok\"
-                                    FROM r3)
-                        SELECT documentdb_core.row_get_bson(r4) AS document
-                        FROM r4".to_string(),
+            list_databases: "SELECT documentdb_api.list_databases($1)".to_string(),
```
jayanta-mondal, line 513: This is potentially a breaking change - if someone has a taken a dependency on the old behavior, the new one might break them. 

So, perhaps the change query needs to behind configuration,.

jayanta-mondal (changes requested):

pg_documentdb_gw/src/postgres/query_catalog.rs
seoc01: The two changes to the output are:
* Adding "sizeOnDisk" and "empty" fields to the output when "nameOnly" is false
* Adding the "totalSize" field 

Do we have guidance on what counts as a "breaking change"? In general as we're developing my opinion is  that adding fields to more closely match the specification for the operation should not be considered breaking, just to avoid a huge number of flags.

For this specific operation I think adding "totalSize" makes sense.

"sizeOnDisk" and "empty" are provided using the existing "list_databases" which sets the values to defaults (0 size, empty false). I would normally prefer to not set the fields rather than have them as dummy values, but since they are coming from another public API changing that would also be potentially breaking.  Personally, I'd rather live with them being dummy values for now.

drk-knght:
@AndrewKhoma 

> can we add more tests to pg_documentdb_gw/documentdb_tests/src/commands/list_databases.rs?

I am not sure which file you are pointing to, please correct if I am wrong is this the `command_list_databases` UDF implementation in `db_stats.c`. If its that, sure we can add few more SQL test cases for that. Is there any specific tests that you think should be tested or in general few more test cases?

drk-knght:
@visridha 

> Could you please also add/replicate tests under pg_documentdb for non distributed test coverage?

Sure wil add that in the next rev cycle.

pg_documentdb_gw/src/postgres/documentdb_data_client.rs
```diff
@@ -588,39 +588,22 @@ impl PgDataClient for DocumentDBDataClient {
     ) -> Result<Response> {
         let (request, request_info, request_tracker) = request_context.get_components();
         // TODO: Handle the case where !nameOnly - the legacy gateway simply returns 0s in the appropriate format
```
AndrewKhoma, line 590: we can remove this comment too

pg_documentdb_gw/src/postgres/query_catalog.rs
AndrewKhoma: no, it's not breaking to add new fields, it's more about the schema change since list_databases could be provided in the recent release. in this case we depend on old api (which was checked-in as part of 1.102 https://github.com/documentdb/documentdb/blob/d0d2317ff6bc39748c05cf1802ad1747abd256ab/pg_documentdb/sql/udfs/metadata/list_databases--0.102-0.sql) so it's safe to do this change

AndrewKhoma (approved):

PR #489, opened 2026-03-10
Copr packaging

xgerman:


xgerman:
Supserseded by https://github.com/alaye-ms/documentdb/blob/users/alaye/coprSpec/packaging/rpm/spec/postgres18-documentdb.spec

PR #498, opened 2026-03-11
fix: Skip sample data init on restart if data already exists (#479)

### LINKED ISSUE #479: DocumentDB-local Container restartfials due to duplicate initialziation of demo data
**Describe the bug**
DocumentDB always fails when the container is restarted.  This is because we auto-populate the sample data during emulator initialization, which fails with a DuplicateID error causing my devcontainer to fail.

**Environment**
- Operating system and version
- PostgreSQL version
- Architecture

**Reproduction Steps**

A code snipped or accessible repository that can reproduce the issue.

**Expected behavior**

**Actual behavior**

**Additional context**

Ritvik-Jayaswal:
Fixes #479

## Problem

When the DocumentDB-local emulator container is restarted with a persisted `--data-path`, the entrypoint always re-runs sample data initialization. Since the sample JS files use `insertMany` with hardcoded `_id` values, this fails with a `DuplicateID` error, crashing the container on restart.

## Fix

- Added a `check_database_has_data` helper that uses `mongosh` to check if a database already contains data.
- Before loading built-in sample data, check if `sampledb` already has data — if so, skip initialization with a log message.
- For custom init data (`--init-data-path`), changed the failure from a hard `exit 1` to a warning + continue, since custom scripts may also fail on restart if data already exists.

The text: "Sample data already exists..." shows in the output to confirm this fix:
image

### DIFF
--- a/scripts/emulator_entrypoint.sh
+++ b/scripts/emulator_entrypoint.sh
@@ -424,13 +424,17 @@ fi
 
 gateway_pid=$! # Capture the PID of the gateway process
 
-# Wait for the gateway to be ready before attempting initialization
+# Wait for the gateway to be ready before attempting initialization.
+# We check /proc/net/tcp6 for a LISTEN socket on the gateway port to avoid
+# making a raw TCP connection that would crash the TLS-enabled gateway.
 echo "Waiting for gateway to be ready..."
+GATEWAY_PORT_HEX=$(printf '%04X' "$DOCUMENTDB_PORT")
 max_attempts=60
 attempt=0
 while [ $attempt -lt $max_attempts ]; do
-    if nc -z localhost "$DOCUMENTDB_PORT"; then
-        echo "Gateway is ready on localhost:$DOCUMENTDB_PORT"
+    if grep -q ":${GATEWAY_PORT_HEX} " /proc/net/tcp6 2>/dev/null && \
+       grep ":${GATEWAY_PORT_HEX} " /proc/net/tcp6 2>/dev/null | grep -q ' 0A '; then
+        echo "Gateway is listening on port $DOCUMENTDB_PORT"
         break
     fi
     echo "Attempt $((attempt + 1))/$max_attempts: Gateway not ready yet, waiting..."
--- a/scripts/emulator_entrypoint.sh
+++ b/scripts/emulator_entrypoint.sh
@@ -444,6 +444,29 @@ if [ $attempt -eq $max_attempts ]; then
     exit 1
 fi
 
+# Helper function to check if a database already has data via mongosh.
+# Returns 0 (true) if the database exists and contains at least one collection with documents.
+check_database_has_data() {
+    local db_name="$1"
+    if command -v mongosh >/dev/null 2>&1; then
+        local doc_count
+        doc_count=$(mongosh "localhost:${DOCUMENTDB_PORT}" \
+            -u "$USERNAME" -p "$PASSWORD" \
+            --authenticationMechanism SCRAM-SHA-256 \
+            --tls --tlsAllowInvalidCertificates \
+            --quiet --eval "
+                use('${db_name}');
+                const cols = db.getCollectionNames();
+                if (cols.length === 0) { print('0'); }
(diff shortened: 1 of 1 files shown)


scripts/emulator_entrypoint.sh
```diff
@@ -444,6 +444,29 @@ if [ $attempt -eq $max_attempts ]; then
...
+        local doc_count
+        doc_count=$(mongosh "localhost:${DOCUMENTDB_PORT}" \
+            -u "$USERNAME" -p "$PASSWORD" \
+            --authenticationMechanism SCRAM-SHA-256 \
+            --tls --tlsAllowInvalidCertificates \
+            --quiet --eval "
+                use('${db_name}');
+                const cols = db.getCollectionNames();
+                if (cols.length === 0) { print('0'); }
+                else { print(db.getCollection(cols[0]).countDocuments({}, { limit: 1 })); }
```
AndrewKhoma: what if col[0] doesn't have documents, but col[1] do?

AndrewKhoma: can we just call dbstats instead?

visridha: Why not just write a marker file saying initialization is done and if it is don't call it again?

guanzhousongmicrosoft (commented):
I would prefer a marker file indicating initialization status

scripts/emulator_entrypoint.sh
Ritvik-Jayaswal: Added marker file for sample data. Also made sample data opt-in with flag --load-sample-data.

guanzhousongmicrosoft:
In the custom init failure path, we now warn and continue, but the marker file is only written on success. With a persisted data dir, that means the same failing custom init will be retried on every restart and keep warning forever. Should this either still fail fast, or persist some state so restart behavior converges?

scripts/emulator_entrypoint.sh
```diff
@@ -445,64 +444,76 @@ if [ $attempt -eq $max_attempts ]; then
...
+            echo ""
+            echo "Connect to your DocumentDB instance and use: use('sampledb')"
+        else
+            echo "Warning: Sample data or initialization script not found"
+            if [ ! -f "$init_script" ]; then
+                echo "  - Missing: $init_script"
+            fi
+            if [ ! -d "$sample_data_path" ]; then
+                echo "  - Missing: $sample_data_path"
+            fi
```
AndrewKhoma, line 518: shouldn't we exit here, since we are expecting to load sample data?

scripts/emulator_entrypoint.sh
```diff
@@ -445,64 +444,76 @@ if [ $attempt -eq $max_attempts ]; then
 fi
 
 # Initialize database with custom data if directory exists and contains JS files
+CUSTOM_INIT_MARKER="$DATA_PATH/.custom_data_initialized"
 custom_data_initialized=false
 if [ -d "$INIT_DATA_PATH" ] && [ "$(ls -A "$INIT_DATA_PATH"/*.js 2>/dev/null)" ]; then
```
AndrewKhoma, line 455: why do we keep loading custom data and not check LOAD_SAMPLE_DATA flag?

Ritvik-Jayaswal: By design, custom data (--init-data-path) and sample data (--load-sample-data) are independent features. A user might use custom data without sample data. I added a comment to clarify this.

scripts/emulator_entrypoint.sh
Ritvik-Jayaswal: Yes, if user explicitly passed --load-sample-data, a missing script/dir is a broken image and should fail hard, added exit here.

Ritvik-Jayaswal:
> In the custom init failure path, we now warn and continue, but the marker file is only written on success. With a persisted data dir, that means the same failing custom init will be retried on every restart and keep warning forever. Should this either still fail fast, or persist some state so restart behavior converges?

Fixed: Now writes marker file even on failure, so restart converges.

xgerman (approved):

alaye-ms:
Stale PR.

Issue #492, opened 2026-03-10
[Feature Request] Make Gateway backward compatible with older Extension versions


v-sathvik:
**Purpose of the feature.**
When the Gateway is updated to use new BSON command signatures but the Extension has not yet been updated, commands like `renameCollection` fail with a SQL error because the new function signature doesn't exist in the older Extension. There is currently no mechanism to handle this Gateway/Extension version mismatch gracefully.

**Describe the solution you'd like**
Add version-based routing in the Gateway:
  1. Parse the Extension version from `topology_bson`
  2. Add a GUC flag `enableBsonPassthroughCommands` in the Extension (default: `false`)
  3. In the Gateway: if the Extension version supports BSON signatures AND GUC is
     enabled → use new BSON path, otherwise fall back to the old scalar param path

This allows a phased rollout — ship the code with GUC off, enable it on internal accounts first to validate, then enable by default on the next release.

**Describe alternatives you've considered**
  Requiring Gateway and Extension to always be updated together (no version mismatch
  allowed). This is simpler but risky — a single missed update would break all affected
  commands with no fallback.

**Additional context**
- When GUC is `false` (default): Gateway calls `rename_collection($1, $2, $3, $4)` —
    works with any Extension version
- When GUC is `true` and Extension version is new enough: Gateway uses the new
    full BSON path 

  Related: PR #468, RFC #440

v-sathvik:
I am working on this.

PR #514, opened 2026-03-12
Add version check to enable backward compatible Gateway-Extension BSO…

v-sathvik:
- Add GUC enableBsonPassthroughCommands (default: off) in Extension
  - Parse Extension version from topology_bson in Gateway
  - Route commands to BSON or legacy path based on GUC + Extension version >= 0.110
  - Add is_bson_passthrough_enabled() test helper for conditional test skipping

Related to #492

### DIFF
--- a/pg_documentdb/src/configs/feature_flag_configs.c
+++ b/pg_documentdb/src/configs/feature_flag_configs.c
@@ -387,6 +387,14 @@ bool EnableDropInvalidIndexesOnReadOnly = DEFAULT_ENABLE_DROP_INDEXES_ON_READ_ON
 #define DEFAULT_INDEX_BUILDS_SCHEDULED_ON_BGWORKER false
 bool IndexBuildsScheduledOnBgWorker = DEFAULT_INDEX_BUILDS_SCHEDULED_ON_BGWORKER;
 
+/*
+ * SECTION: Gateway-Extension compatibility flags
+ */
+
+/* Added in v111, Pending stabilization */
+#define DEFAULT_ENABLE_BSON_PASSTHROUGH_COMMANDS false
+bool EnableBsonPassthroughCommands = DEFAULT_ENABLE_BSON_PASSTHROUGH_COMMANDS;
+
 /* FEATURE FLAGS END */
 
 void
@@ -970,5 +978,14 @@ InitializeFeatureFlagConfigurations(const char *prefix, const char *newGucPrefix
 			"Whether to enable dropping invalid indexes on read only database state."),
 		NULL, &EnableDropInvalidIndexesOnReadOnly,
 		DEFAULT_ENABLE_DROP_INDEXES_ON_READ_ONLY,
+		PGC_USERSET, 0, NULL, NULL, NULL);
+
+	DefineCustomBoolVariable(
+		psprintf("%s.enableBsonPassthroughCommands", newGucPrefix),
+		gettext_noop(
+			"Enables BSON passthrough mode for gateway commands that depend on newer extension versions."),
+		NULL,
+		&EnableBsonPassthroughCommands,
+		DEFAULT_ENABLE_BSON_PASSTHROUGH_COMMANDS,
 		PGC_USERSET, 0, NULL, NULL, NULL);
 }
--- a/pg_documentdb_gw/documentdb_gateway_core/src/configuration/dynamic.rs
+++ b/pg_documentdb_gw/documentdb_gateway_core/src/configuration/dynamic.rs
@@ -14,6 +14,15 @@ use crate::{configuration::Version, postgres::conn_mgmt};
 
 pub const POSTGRES_RECOVERY_KEY: &str = "IsPostgresInRecovery";
 
+/// Parses "major.minor-patch" or "major.minor.patch" format into (major, minor).
+pub fn parse_extension_version(version_str: &str) -> Option<(u32, u32)> {
+    let (major_str, rest) = version_str.split_once('.')?;
+    let major = major_str.parse().ok()?;
+    let minor_str = rest.split(['-', '.']).next()?;
+    let minor = minor_str.parse().ok()?;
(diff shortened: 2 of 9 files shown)


pg_documentdb/src/configs/feature_flag_configs.c
```diff
@@ -971,4 +979,13 @@ InitializeFeatureFlagConfigurations(const char *prefix, const char *newGucPrefix
 		NULL, &EnableDropInvalidIndexesOnReadOnly,
 		DEFAULT_ENABLE_DROP_INDEXES_ON_READ_ONLY,
 		PGC_USERSET, 0, NULL, NULL, NULL);
+
+	DefineCustomBoolVariable(
+		psprintf("%s.enableBsonPassthroughCommands", newGucPrefix),
+		gettext_noop(
+			"Enables BSON passthrough mode for gateway commands that depend on newer extension versions."),
```
AndrewKhoma, line 986: need @safern and @visridha opinion on this, where did we landed at the end of the convo? should we have a gateway specific space for this GUCs, right?

pg_documentdb_gw/documentdb_gateway_core/src/configuration/dynamic.rs
```diff
@@ -14,6 +14,15 @@ use crate::{configuration::Version, postgres::conn_mgmt};
 
 pub const POSTGRES_RECOVERY_KEY: &str = "IsPostgresInRecovery";
 
+/// Parses "major.minor-patch" or "major.minor.patch" format into (major, minor).
+pub fn parse_extension_version(version_str: &str) -> Option<(u32, u32)> {
```
AndrewKhoma, line 18: can we have a version struct, please? and also let's have a patch here too

pg_documentdb_gw/documentdb_gateway_core/src/configuration/dynamic.rs
```diff
@@ -144,4 +153,32 @@ pub trait DynamicConfiguration: Send + Sync + Debug {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         write!(f, "")
     }
+
+    /// Returns true if the GUC enableBsonPassthroughCommands is ON and the Extension
+    /// version is >= 0.110 (the version where BSON command signatures were introduced).
+    fn extension_supports_bson_passthrough(&self) -> bool {
```
AndrewKhoma, line 159: let's not overcrowd this method, instead have this call in `process_rename_collection` in `data_descriptions.rs`, so we don't expand the process.rs more than needed. there you can do the check of dynamic config and check for version

pg_documentdb_gw/documentdb_gateway_core/src/processor/data_description.rs
```diff
@@ -118,6 +118,71 @@ pub async fn process_rename_collection(
...
+        .execute_rename_collection_legacy(
+            request_context,
+            source_db,
+            source_coll,
+            target_coll,
+            drop_target,
+            connection_context,
+        )
+        .await?;
+    Ok(Response::ok())
```
AndrewKhoma, line 183: nit: we can return this in pg_data_client response

pg_documentdb_gw/documentdb_gateway_core/src/processor/process.rs
```diff
@@ -291,12 +291,21 @@ pub async fn process_request(
...
+                    )
+                    .await
+                } else {
+                    data_description::process_rename_collection_legacy(
+                        request_context,
+                        connection_context,
+                        pg_data_client,
+                    )
+                    .await
+                }
```
AndrewKhoma, line 308: let's move this to be handled inside of the data_description::process_rename_collection?

pg_documentdb_gw/documentdb_tests/src/test_setup/postgres.rs
```diff
@@ -61,6 +61,30 @@ pub fn get_pool_manager() -> Arc<PoolManager> {
     }))
 }
 
+pub async fn is_bson_passthrough_enabled() -> bool {
+    let pool_manager = get_pool_manager();
+    let conn = match pool_manager.system_requests_connection().await {
+        Ok(c) => c,
+        Err(_) => return false,
+    };
```
AndrewKhoma, line 69: please, don't hold system connection longer than needed, i highly recommend have this call inside of the clause that would drop the conn as soon as you're out of it

pg_documentdb_gw/documentdb_tests/tests/rename_collection_tests.rs
```diff
@@ -49,6 +52,11 @@ async fn validate_rename_collection_not_found_error() -> Result<(), Error> {
 
 #[tokio::test]
 async fn validate_rename_collection_unknown_field_error() -> Result<(), Error> {
+    if !postgres::is_bson_passthrough_enabled().await {
+        // Unknown field validation is only enforced on the BSON path (GUC=on).
+        // On the legacy path, unknown fields are silently ignored by design.
+        return Ok(());
+    }
```
AndrewKhoma, line 59: can you mock the version and the guc to see that the switch works?

pg_documentdb/src/configs/feature_flag_configs.c
safern: IMHO I think we should have a gateway feature flags file if there are features that are only used by the gateway?  Since feature flags are mostly used for backend features where the backend code takes decisions based on it. 

The question I'd like us to discuss here is, do we really need a GUC for this one? Also what will be the path to enable by default cause I see default is off? The command was new and not supported in docdb so I think we should just have it enabled by default and have the version check.

xgerman: I think having the configuration in a central place (GUC) is helpful - we had issues with files before + it introduces a second configuration schema.

PR #517, opened 2026-03-13
Add BSON signature for drop_collection and drop_database

v-sathvik:
- Add BSON signature for drop_collection in Extension
  - Add BSON signature for drop_database in Extension
  - Update Gateway to pass full BSON for drop_collection and drop_database                                                                                                    
  - Add drop_collection and drop_database BSON regression tests
  - Add Gateway integration tests for both commands                                                                                                                           
                  
  #### Summary:
  Currently the Gateway extracted scalar fields (db name, collection name) from the client's
  OP_MSG BSON and passed them individually to the Extension. This discarded session info
  (lsid), writeConcern, and other fields.

  This change adds new BSON-accepting signatures `drop_collection(bson)` and
  `drop_database(bson)` so the Gateway passes the complete client BSON to the Extension.
  The Extension parses the BSON internally to extract the needed fields. The legacy text
  signatures are preserved for backward compatibility.

  Gateway routing uses `extension_supports_bson_passthrough()` to switch between BSON
  and legacy paths — depends on PR #514 (version-check infrastructure) merging first.

  Related to #440, #492, #514


pg_documentdb/src/commands/drop_collection.c
```diff
@@ -190,7 +249,68 @@ command_drop_collection(PG_FUNCTION_ARGS)
...
+	args.hasUuid = !PG_ARGISNULL(1);
+	if (args.hasUuid)
+	{
+		args.collectionUuid = PG_GETARG_DATUM(1);
+	}
+	args.trackChanges = PG_ARGISNULL(2) ? true : PG_GETARG_BOOL(2);
+
+	ThrowIfServerOrTransactionReadOnly();
+
+	PG_RETURN_BOOL(ValidateAndExecuteDrop(&args, commandSpec, NULL));
```
visridha: prefer to return a bson with OK: 1

pg_documentdb_gw/documentdb_gateway_core/src/configuration/dynamic.rs
```diff
@@ -29,6 +29,12 @@ pub trait DynamicConfiguration: Send + Sync + Debug {
     // Needed to downcast to concrete type
     fn as_any(&self) -> &dyn std::any::Any;
 
+    /// Temporary stub . Remove after PR#514 merges —
```
AndrewKhoma, line 32: so should we take a dependency on that PR and wait for it to merge prior to merging this one?

pg_documentdb/sql/udfs/schema_mgmt/drop_collection--0.111-0.sql
```diff
@@ -0,0 +1,20 @@
+/* with BSON */
+CREATE OR REPLACE FUNCTION __API_SCHEMA_V2__.drop_collection(
```
AndrewKhoma, line 2: should we have `DROP FUNCTION IF EXISTS __API_SCHEMA_V2__.drop_collection CASCADE;`?

pg_documentdb/sql/udfs/schema_mgmt/drop_collection--latest.sql
```diff
@@ -25,3 +25,24 @@ LANGUAGE C
 AS 'MODULE_PATHNAME', $function$command_drop_collection$function$;
 COMMENT ON FUNCTION documentdb_api_v2.drop_collection(text,text,__CORE_SCHEMA_V2__.bson,uuid,bool)
     IS 'drop a collection';
+
+/* with BSON */
+CREATE OR REPLACE FUNCTION __API_SCHEMA_V2__.drop_collection(
```
AndrewKhoma, line 30: previous sql functions had `DROP FUNCTION IF EXISTS documentdb_api_v2.drop_collection CASCADE;` before their definition, should we have the same here?

pg_documentdb/src/commands/drop_collection.c
```diff
@@ -61,53 +116,58 @@ command_drop_collection(PG_FUNCTION_ARGS)
...
+		DistributedRunCommandResult result = RunCommandOnMetadataCoordinator(
+			dropQuery->data);
 
 		if (!result.success)
 		{
 			ereport(ERROR, (errcode(ERRCODE_DOCUMENTDB_INTERNALERROR),
 							errmsg(
 								"Internal error dropping collection in metadata coordinator"),
-							errdetail_log(
-								"Internal error dropping collection in metadata coordinator %s",
```
AndrewKhoma, line 96: why do we change the error message here?

pg_documentdb/src/commands/drop_database.c
```diff
@@ -0,0 +1,121 @@
...
+static void
+ExecuteDropDatabase(const char *databaseName)
+{
+	/* Delegate to the existing drop_database(text) which already handles
+	 * querying the collections catalog and dropping each collection. */
+	StringInfo dropQuery = makeStringInfo();
+	appendStringInfo(dropQuery,
+					 "SELECT %s.drop_database(%s)",
+					 ApiSchemaName,
+					 quote_literal_cstr(databaseName));
```
AndrewKhoma, line 87: why are we calling this select function if we can drop collections directly from the backend? feels like some slowdown. @LordParag what do you think?

pg_documentdb/src/test/regress/basic_schedule
```diff
@@ -16,6 +16,8 @@ test: bson_aggregation_pipeline_tests_merge_objects_group bson_aggregation_curso
 test: bson_aggregation_pipeline_tests_stddevpopsamp_group readonly_transaction_tests bson_orderby_composite_filtering_tests bson_composite_index_tests_wildcard_tests
 test: commands_create_indexes_background_bgworker commands_create_view_tests bson_expr_index_pushdown_tests
 test: collection_management!PG18_OR_HIGHER! bson_aggregation_cursor_tests_txn bson_composite_index_tests_multi_key
+test: drop_collection_bson_spec
+test: drop_database_bson_spec
```
AndrewKhoma, line 20: can we merge these together?

pg_documentdb_gw/documentdb_gateway_core/src/processor/process.rs
```diff
@@ -121,22 +121,42 @@ pub async fn process_request(
...
             }
             RequestType::Drop => {
-                data_description::process_drop_collection(
-                    request_context,
-                    connection_context,
-                    &dynamic_config,
-                    pg_data_client,
-                )
-                .await
+                if dynamic_config.extension_supports_bson_passthrough() {
```
AndrewKhoma, line 124: let's have this logic inside data_description, rather then here so we don't bloat already big process.rs

pg_documentdb_gw/documentdb_gateway_core/src/processor/process.rs
```diff
@@ -121,22 +121,42 @@ pub async fn process_request(
...
             }
             RequestType::DropDatabase => {
-                data_description::process_drop_database(
-                    request_context,
-                    connection_context,
-                    &dynamic_config,
-                    pg_data_client,
-                )
-                .await
+                if dynamic_config.extension_supports_bson_passthrough() {
```
AndrewKhoma, line 143: same comment as above

pg_documentdb_gw/documentdb_tests/src/commands/drop_collection.rs
```diff
@@ -0,0 +1,73 @@
...
+const ERR_UNKNOWN_BSON_FIELD: i32 = 40415; // ERRCODE_DOCUMENTDB_UNKNOWNBSONFIELD (M0088)
+
+pub async fn validate_drop_collection_basic(db: &Database) -> Result<(), Error> {
+    db.collection::<Document>("test_coll")
+        .insert_one(doc! { "_id": 1, "a": 1 })
+        .await?;
+
+    let result = db.run_command(doc! { "drop": "test_coll" }).await?;
+
+    assert_eq!(result.get_f64("ok").unwrap(), 1.0);
```
AndrewKhoma, line 26: nit, can we not unwrap, but rather check the option?

pg_documentdb_gw/documentdb_tests/src/commands/drop_collection.rs
```diff
@@ -0,0 +1,73 @@
+/*-------------------------------------------------------------------------
```
AndrewKhoma, line 1: can we have some test that actually validates the switch between different versions?

pg_documentdb_gw/documentdb_tests/src/commands/drop_database.rs
```diff
@@ -0,0 +1,49 @@
+/*-------------------------------------------------------------------------
```
AndrewKhoma, line 1: can we have a test to validate the work of the switch?

pg_documentdb/src/commands/drop_database.c
paragikjain: Yeah, I think it’s better to perform the operation directly in the backend, as it can avoid unnecessary query planning and execution overhead. The only requirement is to fetch all collections for the database from the catalog, after which we can directly invoke the C function to drop each collection.

pg_documentdb/sql/udfs/schema_mgmt/drop_collection--0.111-0.sql
safern: CREATE OR REPLACE is handling this right? REPLACE will make it safe if the function already exists.

pg_documentdb/sql/udfs/schema_mgmt/drop_collection--latest.sql
safern: DROP is usually only needed when the comments or arguments change, if they are identical the REPLACE should handle it.

paragikjain:
hi @v-sathvik 

 just following up on this contribution. We reviewed it about a month ago and had shared some feedback. Could you please take a look and let us know if you plan to address the comments? Happy to help if you need any clarification. Thanks!

PR #524, opened 2026-03-13
Add BSON signature support for shard_collection and reshard_collection

v-sathvik:
- Update Gateway to pass full BSON for shard_collection and reshard_collection
- Add shard_collection BSON regression tests
- Add Gateway integration tests for shard_collection and reshard_collection

#### Summary:
The BSON signatures for `shard_collection(bson)` and `reshard_collection(bson)` have existed in the Extension since v0.24 - the C function already handles both the legacy 4-param path and the BSON path via `PG_NARGS()` check.

This change updates the Gateway to pass the complete client BSON directly instead of extracting scalar fields (db, collection, key, reshard). This preserves session info (lsid), writeConcern, and other fields that were previously discarded.

Since the BSON signature has always existed, no version-check routing is needed the BSON path is safe to call against any Extension version.

Related to #440, #514


pg_documentdb_gw/documentdb_gateway_core/src/processor/data_description.rs
```diff
@@ -121,27 +121,22 @@ pub async fn process_rename_collection(
...
-    let (db, collection) =
-        protocol::extract_database_and_collection_names(collection_path.as_str())?;
-    let key = request_context
-        .payload
-        .document()
-        .get_document("key")
-        .map_err(DocumentDBError::parse_failure())?;
+    pg_data_client
+        .execute_shard_collection(request_context, connection_context)
+        .await?;
```
AndrewKhoma: we can just return the data_client response without ? and Ok(response::ok())

pg_documentdb_gw/documentdb_gateway_core/src/processor/data_description.rs
```diff
@@ -121,27 +121,22 @@ pub async fn process_rename_collection(
...
     pg_data_client
-        .execute_shard_collection(
-            request_context,
-            db,
-            collection,
-            key,
-            reshard,
-            connection_context,
-        )
+        .execute_reshard_collection(request_context, connection_context)
```
AndrewKhoma, line 137: same comment from above applies to this method

pg_documentdb_gw/documentdb_tests/src/commands/shard_collection.rs
```diff
@@ -0,0 +1,62 @@
...
+        "The key parameter is required",
+    )
+    .await;
+}
+
+pub async fn validate_reshard_collection_missing_key_error(db: &Database) {
+    db.collection::<Document>("test_coll")
+        .insert_one(doc! { "_id": 1 })
+        .await
+        .unwrap();
```
AndrewKhoma: can we not unwrap, please?

AndrewKhoma:
overall - LGTM

pg_documentdb_gw/documentdb_tests/src/commands/shard_collection.rs
v-sathvik: Ack.

pg_documentdb/src/test/regress/sql/shard_collection_bson_spec.sql
```diff
@@ -0,0 +1,17 @@
+SET search_path TO documentdb_api, documentdb_api_catalog, documentdb_core;
```
safern, line 1: nit usually the test files end with _tests

pg_documentdb_gw/documentdb_tests/src/commands/shard_collection.rs
```diff
@@ -0,0 +1,65 @@
...
+    commands::execute_command_and_validate_error(
+        db,
+        doc! {
+            "reshardCollection": format!("{}.test_coll", db.name()),
+        },
+        ERR_FAILED_TO_PARSE,
+        "The key parameter is required",
+    )
+    .await;
+}
```
AndrewKhoma, line 65: in `data_descriptions.rs` we are removing the call of `extract_database_and_collection_names` expecting that backend will catch the parsing error for us, but I don't see any test for gw to confirm that we're still preserving the same functionality as we would before (we need to return BadValue)

PR #528, opened 2026-03-17
Add RBAC impl proposal

seoc01:
DocumentDB RBAC Implementation Overview


rfcs/0008-rbac-impl-design.md
```diff
@@ -0,0 +1,273 @@
...
+                  }
+```
+
+The users table references the Postgres role associated with the user, and stores the relevant user data in a bson object. The user_data object will store the roles associated with the user in a list of database/role name pairs (e.g. `roles: [{"db":"test", "role":"write"}, {"db":"admin", "role":"readAnyDatabase"}]`). Additional fields associated with the user can be added in the future, such as custom data, command, authenticationRestrictions (see: https://www.mongodb.com/docs/manual/reference/command/usersInfo/#output)
+
+### Roles Table
+
+```
+CREATE TABLE IF NOT EXISTS documentdb_api_catalog.roles (
+    role_document documentdb_core.bson NOT NULL
```
seoc01: Create matching PG role? To be followed up on.

rfcs/0008-rbac-impl-design.md
```diff
@@ -0,0 +1,425 @@
...
+        "collMod", "collStats", "dbStats",
+        "listCollections", "listIndexes"
+      ]
+    }
+  ],
+  "roles": []
+}
+```
+
+Built-in roles have their definitions stored in the code, and are only added to the roles table as needed.
```
DeepakSPemmaraju, line 73: Just to clarify this would need to be in the pg_roles table, we're just not adding it to the custom roles table? Also would this mean for built-in roles pg_roles table is the source of truth?

seoc01: They would be added to both the custom DocDb roles table, and to the pg_roles table, the pg_roles table would only have information about the insert/update/delete privileges, the DocDb roles table would store all of the other privileges.

rfcs/0008-rbac-impl-design.md
```diff
@@ -0,0 +1,425 @@
...
+               ],
+               comment: "..."}
+```
+
+The users table references the Postgres role associated with the user, and stores the relevant user data in a bson object. The user_data object will store the roles associated with the user in a list of database/role name pairs (e.g. `roles``:`` ``[{"db":``"test", "role":``"write"``},`` ``{"db":``"admin", "role":``"readAnyDatabase"``}]`). Additional fields associated with the user can be added in the future, such as custom data, command, authenticationRestrictions (see: https://www.mongodb.com/docs/manual/reference/command/usersInfo/#output)
+
+### Roles Table
+
+```
+CREATE TABLE IF NOT EXISTS documentdb_api_catalog.roles (
```
DeepakSPemmaraju, line 35: We already have a roles table that was checked in a few months ago, can you take a look at that

rfcs/0008-rbac-impl-design.md
```diff
@@ -0,0 +1,425 @@
...
+
+## DocumentDB Roles
+
+DocumentDB has 3 types of roles: 
+
+* Static Built-in roles, these are standard built in roles, like “dbAdmin” with are associated with the “admin” database, there is only one version of each of these roles
+* Dynamic Built-in roles, these are built in roles that are associated with a database, e.g. `{“db”:“test”, “role”: “write”}`. These provide built in privileges based on the named database where they are assigned. For each of these role definitions there is theoretically one per database the customer creates. 
+* Custom Role, these are roles the customer creates, that are associated with the database the customer creates them in (but can have privileges outside that database), these can also inherit from other roles. 
+
+To support all types of roles, we will only create roles in the DocumentDb roles table as roles are needed, e.g. built-in role used or custom role created. The definitions for built-in roles will live in the code.
```
DeepakSPemmaraju, line 288: I'm not sure if we already discussed this, but we want to maintain parity between pg_roles tables and the custom roles table right? Essentially every role in PG will also exist in the custom roles table and vice versa? Also are we doing the same for users table? So every user in the users table also exists as a role with login privileges in pg_roles table?

rfcs/0008-rbac-impl-design.md
```diff
@@ -0,0 +1,425 @@
...
+The following “critical” MongoDb actions are mapped to Postgres roles, All other privileges will be enforced by the DocumentDb extension.
+
+find → SELECT
+insert → INSERT
+remove → DELETE
+update → UPDATE
+
+These actions can be assigned at the database+collection level, the database, the cluster level, and associated with a collection name regardless of database. 
+
+**Note**: We do not create a combined “write” role to maintain consistency between built-in and custom roles.
```
DeepakSPemmaraju, line 303: Can you elaborate what you mean by this?

rfcs/0008-rbac-impl-design.md
seoc01: I'll take a look, thanks

rfcs/0008-rbac-impl-design.md
seoc01: Yes, we would keep a 1:1 mapping

rfcs/0008-rbac-impl-design.md
seoc01: I assume you mean for the note (**Note**: We do not create a combined “write” role to maintain consistency between built-in and custom roles.)

I was thinking that instead of granting a general "write" role for a collection, even though it would be the most common use case, we would grant the "update" / "delete" / "insert" privileges separately

Issue #510, opened 2026-03-11
Conflicting update operators on same field path accepted instead of returning an error


ramgokul-14:
**Behavior:** MongoDB rejects conflicting operators on same field, DocumentDB allows them with unexpected results.

**Reproduction:**
```javascript
// Insert document
db.collection.insertOne({foo: 'bar'})

// Upsert with conflicting operators
db.collection.updateOne(
  {foo: 'bar'},
  {
    $mul: {y: -21621},
    $setOnInsert: {y: true}
  },
  {upsert: true}
)
```

**MongoDB Result:** Error
```
Error: Updating the path 'y' would create a conflict at 'y'
```

**DocumentDB Result:** No error, y set to 0
```json
{"_id": ObjectId("..."), "foo": "bar", "y": 0}
```

xgerman:
Verified fixed against documentdb-local:latest (v0.109-0). Live test: conflicting $set+$inc on same path correctly throws conflict error.

xgerman:
Reopening: re-ran the exact reproduction steps from this issue against documentdb-local:latest (v0.109-0) and the bug still reproduces. The earlier closure was based on an agent test that used a simplified setup which did not accurately reproduce the issue. Detailed repro output will follow.

srwshkrshnn:
I can work on this, thanks!

PR #540, opened 2026-03-23
MJ: always check for conflicts on insertUpdate even with no-ops; upda…

mjtung:
## Github Issue: 
https://github.com/documentdb/documentdb/issues/510

### Problem

When an update operation combined `$setOnInsert` with another operator that targeted the same field (e.g., `$set`, `$inc`, `$mul`), the conflict was silently ignored on non-upsert updates. This is because `HandleSetOnInsertUpdateTree` returned early when `isUpsert` was `false`, skipping the path conflict check entirely.

For example, this should always error but previously did not when the document existed:

```js
db.col.update({ foo: "bar" }, { $mul: { y: -21621 }, $setOnInsert: { y: true } })
```

### Root Cause

In `bson_update_operators_workflow.c`, `HandleSetOnInsertUpdateTree` short-circuited with an early `return` for non-upserts before walking the update tree. Path conflict detection happens during tree construction, so the conflicting path was never registered.

### Fix

Remove the early `return` guard so that `$setOnInsert` paths are always added to the update tree. This ensures path conflict detection fires correctly in all cases. The `isUpsert` flag is still forwarded to `HandleBasicUpdateTree`, which uses it to decide whether to actually apply the `$setOnInsert` values.

### Tests

Added regression tests covering:
- `$mul` + `$setOnInsert` on the same field (the originally reported bug)
- `$set` + `$setOnInsert` on the same field
- `$setOnInsert` + `$inc` on the same field (reverse operator order)
- The same conflict on a true upsert (empty source document)

All cases now correctly return:
```
ERROR: Modifying the path 'y' will result in a conflict occurring at 'y'
```

### DIFF
--- a/pg_documentdb/src/test/regress/expected/bson_update_document_tests.out
+++ b/pg_documentdb/src/test/regress/expected/bson_update_document_tests.out
@@ -1877,6 +1877,19 @@ SELECT documentdb_api_internal.bson_update_document('{"data": "data"}', '{ "": {
 ERROR:  Modifying the path 'a.b.c.d.e.f' will result in a conflict occurring at 'a.b.c.d.e'
 SELECT documentdb_api_internal.bson_update_document('{"data": "data"}', '{ "": { "$set": { "a.b.c.d.e": 1 }, "$inc": {"a.b": 1 } } }', '{}');
 ERROR:  Modifying the path 'a.b' will result in a conflict occurring at 'a.b'
+-- test conflicting update operator paths involving $setOnInsert
+-- $mul + $setOnInsert on same field (reported bug: should error regardless of upsert)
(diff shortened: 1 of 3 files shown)


LordParag (commented):
Hello @mjtung 

Thanks for the fix—LGTM overall. I did notice one test failing:
pg_documentdb_distributed/src/test/regress/expected/bson_update_document_tests.out.

This seems to be because it previously failed with a different error, but now it’s also failing due to a conflict, so the expected output no longer matches. Could you please update the expected output file accordingly? Also, it would be great to add a new test case that still triggers the original error, so we don’t lose coverage.

shuaitian-git:
Hi @mjtung thanks for the contribution here, can you help update this PR and fix the test mentioned above by Parag? Thanks!

PR #588, opened 2026-04-17
Updating CODEOWNERS

tsharp:
- Update CODEOWNERS for Gateway Related Code
- Update CODEOWNERS for Build / Pipeline
- Update CODEOWNERS for /rfcs/

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -2,6 +2,7 @@
 
 /.pipelines/ @safern @visridha @shuaitian-git @lichoil @xgerman @tsharp
 /.github/workflows/ @safern @visridha @shuaitian-git @lichoil @xgerman @tsharp
+/.devcontainer/ @safern @visridha @xgerman @tsharp @documentdb/documentdb-maintainers
 /packaging/ @safern @visridha @shuaitian-git @lichoil @fatmelon @xgerman
 
 *.csv @documentdb/documentdb-maintainers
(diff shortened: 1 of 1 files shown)


xgerman (approved):

CODEOWNERS
```diff
@@ -1,12 +1,20 @@
...
 /internal/ @documentdb/documentdb-maintainers
 /pg_documentdb/ @documentdb/documentdb-maintainers
-/pg_documentdb_core/ @documentdb/documentdb-maintainers
\ No newline at end of file
+/pg_documentdb_core/ @documentdb/documentdb-maintainers
+
+/rust-toolchain.toml @documentdb/gateway-owners
+/pg_documentdb_gw_host/ @documentdb/gateway-owners
+/pg_documentdb_gw/ @documentdb/gateway-owners
+/rfcs/ @documentdb/gateway-owners
```
visridha: Why does all RFC go to gateway owners?

CODEOWNERS
```diff
@@ -1,12 +1,20 @@
...
+/.devcontainer/ @documentdb/gateway-owners
```
visridha: Why do gateway owners pick up devcontainer?

CODEOWNERS
```diff
@@ -1,12 +1,20 @@
...
+/documentdb_errors.csv @documentdb/gateway-owners
```
visridha: why are these on gateway owners?

visridha (changes requested):

CODEOWNERS
tsharp: Good question - in this case I do believe that a shared ownership would likely be better.

tsharp: This as been adjusted for dual-ownership until we can determine a more granular model.

CODEOWNERS
tsharp: More of how it's currently maintained than a .devcontainer specific ownership - I've adjusted this to be more in-line with how the pipelines are currently owned.

CODEOWNERS
tsharp: @visridha - for the csv files - you're correct - this is too permissive and was originally based on this here -> https://github.com/documentdb/documentdb/blob/ac2fa2f4ae0024515bb325274c9574fad8cc27e7/pg_documentdb_gw/documentdb_macros/src/lib.rs#L31

I've removed those permissions and scoped the host and toolchain toml to be better reflective of dual ownership here.

Issue #563, opened 2026-04-07
Ship scripts and sample-data in gateway packages (DEB + RPM)


guanzhousongmicrosoft:
## Parent Issue

Part of #561 (Make DocumentDB usable from package install)

## Problem

The gateway package currently ships only the binary, config, and systemd unit. The `documentdb-setup` CLI (#562) and the Docker emulator entrypoint both depend on helper scripts (`utils.sh`, `start_oss_server.sh`, `emulator_entrypoint.sh`, etc.) and sample data files. Without these, neither the setup CLI nor the Docker image can function from package-installed paths.

## Solution

Include the following in gateway packages at standardized paths:

### Scripts (`/usr/share/documentdb/scripts/`)
- `documentdb-setup.sh` -> installed as `/usr/bin/documentdb-setup` (755)
- `utils.sh`
- `start_oss_server.sh`
- `build_and_start_gateway.sh`
- `emulator_entrypoint.sh`
- `init_documentdb_data.sh`
- `setup_psqlrc.sh`

### Sample Data (`/usr/share/documentdb/sample-data/`)
- `01-users.js`
- `02-products.js`
- `03-orders.js`
- `04-analytics.js`
- `README.md`

### Changes

**DEB** (`pg_documentdb_gw/documentdb_gateway/Cargo.toml`):
- Add 13 asset entries under `[package.metadata.deb]`
- Add `depends = ["jq"]` (required by `documentdb-setup`)
- Update `maintainer-scripts/postinst` with quick-start hint

**RPM** (`packaging/rpm/spec/documentdb_gateway.spec`):
- Add `Requires: jq`
- Add `%install` commands to copy scripts + sample-data
- Add `%files` entries for the new paths

### Runtime Dependency

`jq` is required by `documentdb-setup` for safe JSON construction (SQL injection prevention via `jq -n --arg`). Added as a package dependency so `dpkg -i` / `rpm -i` will flag it.

Issue #612, opened 2026-05-18
documentdb-local restart loop on second boot: sample-data seed re-runs and fails with DuplicateKey (regression of #479)


xgerman:
**Describe the bug**

`documentdb-local:latest` enters a permanent restart loop on the second container boot whenever a persistent data volume is used. The entrypoint unconditionally re-runs the bundled sample-data load (`/home/documentdb/gateway/sample-data/01-users.js` …) on every start, which `insertMany`s rows whose `_id`s already exist from the first boot. The bulk write fails with a duplicate-key error, `init_documentdb_data.sh` returns non-zero, and `emulator_entrypoint.sh` does `exit 1` (line ~469), so any orchestrator with a restart policy loops forever.

This is a regression (or an incomplete close) of:
- #479 — closed 2026-04-15 without the fix landing
- #399 — closed as duplicate of #479
- #498 — the actual fix PR, still **open and unmerged** as of today

Even with `--skip-init-data` available as a workaround, the default behavior should not crash. Re-running an idempotent seed against pre-seeded data is a normal and expected operation for any container with a persistent volume.

**Reproduction Steps**

```yaml
# compose.yml
services:
  documentdb:
    image: ghcr.io/documentdb/documentdb/documentdb-local:latest
    command: ["--username", "localadmin", "--password", "Admin100"]
    ports: ["10260:10260"]
    volumes:
      - documentdb-data:/data
    restart: unless-stopped
volumes:
  documentdb-data:
```

```
docker compose up -d                # first boot: succeeds, seeds sampledb
docker compose restart documentdb   # second boot: restart loop
docker compose logs documentdb | tail
```

**Actual output (truncated)**

```
Initializing users collection...
MongoBulkWriteError: Duplicate key violation on the requested collection: Index '_id_'
Error: Failed to execute: 01-users.js
This indicates invalid JavaScript syntax or operation error.
Error: Failed to process custom initialization data
Error: Sample data initialization failed
```

…then `exit 1` from `emulator_entrypoint.sh:469`, container restarts, repeat indefinitely.

**Expected behavior**

The entrypoint should either:

1. Detect that the sample collections (`sampledb.users` / `products` / `orders` / `analytics`) already exist and skip the load, **or**
2. Make the seed scripts idempotent (e.g. `insertMany(..., { ordered: false })` and tolerate `DuplicateKey` (E11000), or upsert by `_id`), **or**
[...]

Issue #564, opened 2026-04-07
Simplify Docker images to use package-installed script paths


guanzhousongmicrosoft:
## Parent Issue

Part of #561 (Make DocumentDB usable from package install)

## Problem

Docker images currently `COPY` scripts from the build context into the image at build time, duplicating what the gateway package already installs. This creates two different code paths -- the Docker path and the package path -- that can drift apart. It also means the Docker image does not exercise the same file layout that package users get.

## Solution

Since the gateway `.deb` now ships scripts at `/usr/share/documentdb/scripts/` and sample data at `/usr/share/documentdb/sample-data/` (#563), Docker images should use those package-installed paths directly.

### Changes

1. **`Dockerfile_documentdb_local`** (main Docker image):
   - Remove `COPY scripts/` lines
   - Use `/usr/share/documentdb/scripts/` as `WORKDIR` and entrypoint
   - Add `jq` to package install (runtime dependency)

2. **`emulator_entrypoint.sh`** (Docker entrypoint):
   - Add dual-path resolution: check `/usr/share/documentdb/scripts` first, fall back to legacy `/home/documentdb/gateway/scripts` for backward compatibility
   - Use `mktemp` for runtime config copy (instead of predictable `/tmp` path)
   - Proper `trap` cleanup for temp files

3. **Test Dockerfiles** (DEB + RHEL-8 + RHEL-9 gateway tests):
   - Same migration: remove `COPY` lines, use package paths
   - Add `jq` install where needed

4. **`test-gateway-install-entrypoint.sh`**:
   - Update script path to `/usr/share/documentdb/scripts/`

### Benefit

Single source of truth for script paths. Docker images and bare-metal installs use identical file layouts. Any script change in the repo automatically flows through packages to both Docker and bare-metal users.

sthapa:
Looking at making this change

sthapa:
yes, still working on it

Issue #562, opened 2026-04-07
Create documentdb-setup CLI tool for one-command setup


guanzhousongmicrosoft:
## Parent Issue

Part of #561 (Make DocumentDB usable from package install)

## Summary

Create a `documentdb-setup` bash script that automates the full DocumentDB setup in a single command. This is the core deliverable that bridges the gap between package installation and a working DocumentDB instance.

## What It Does

The script performs 7 steps in sequence:

1. **Detect PostgreSQL** -- auto-detects PG version and binary paths (Debian/RHEL layout)
2. **Initialize cluster** -- runs `initdb`, creates data directory, starts PG on port 9712
3. **Configure PostgreSQL** -- writes drop-in `conf.d/documentdb.conf` with all required GUCs, adds HBA entries for TCP localhost, sets `listen_addresses`
4. **Create extensions** -- `CREATE EXTENSION documentdb CASCADE` (+ extended RUM when available)
5. **Create users** -- creates `documentdb` PG superuser role (gateway system user) + MongoDB-compatible application user via `documentdb_api.create_user()`
6. **Configure gateway** -- updates `SetupConfiguration.json` with correct ports
7. **Start gateway** -- enables systemd service (or manual start without systemd)

## Design Decisions

- **Always requires root** -- simpler than split root/non-root flow
- **Drop-in conf.d** -- reversible, does not clobber `postgresql.conf`
- **Merges shared_preload_libraries** -- reads existing config, deduplicates
- **SQL injection prevention** -- uses `jq` for JSON construction + psql variables with heredoc (not `-c`) + explicit `::documentdb_core.bson` cast
- **Behavioral parity with Docker** -- same GUCs as emulator entrypoint (`indexBuildsScheduledOnBgWorker=false`, `ssl=off`, extended RUM GUCs, etc.)
- **Idempotent** -- safe to re-run (skips existing cluster, user, extensions)
- **`--skip-pg-init`** -- supports existing PostgreSQL clusters
- **Port collision detection** -- verifies running instance data_directory matches expected
- **Gateway readiness check** -- 60s timeout, fatal on failure

## CLI Interface

```
Usage: documentdb-setup --username <USER> --password <PASS> [OPTIONS]

Required:
  --username <USER>       MongoDB-compatible username to create
  --password <PASS>       Password for the user

Options:
  --pg-version <VER>      PostgreSQL version (auto-detected if not specified)
  --pg-port <PORT>        PostgreSQL port (default: 9712)
[...]

PR #631, opened 2026-06-11
Adds scripts and sample data files to debian and rhel packages

sthapa:
Addresses issue #563  - tested this with rhel8, rhel9, deb12, and ubuntu24.04 and pg18 so I'm pretty sure the other items on the test matrix will work

### DIFF
--- a/packaging/rpm/spec/documentdb.spec
+++ b/packaging/rpm/spec/documentdb.spec
@@ -31,6 +31,7 @@ Requires:       (pgvector_%{pg_version} or percona-pgvector_%{pg_version})
 Requires:       pg_cron_%{pg_version}
 Requires:       postgis36_%{pg_version}
 Requires:       rum_%{pg_version}
+Requires:       jq
 # Libbson is now bundled, so no runtime Requires for it.
 # pcre2 is statically linked.
 # libbid.a is bundled.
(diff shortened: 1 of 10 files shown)


guanzhousongmicrosoft:
Thanks @sthapa — and especially for testing across rhel8/9, deb12, and ubuntu24.04 on pg18; that coverage is really helpful. 

This is packaging-only (Makefiles + spec/control + the jq dependency) and doesn't change the behavior of the scripts themselves, so it stands cleanly on its own. We'd like to take it — let's get it through review and merged.
 
For full transparency: we have a separate change in flight that hardens the init-script behavior (emulator_entrypoint.sh / init_documentdb_data.sh) to fix the #612 restart-loop. It edits the contents of those two files but none of the files in your PR, so the two are orthogonal and won't conflict — we'll just land that on top. Nothing for you to do, and no need to wait on us.
 
Thanks again for jumping on this!

scripts/Makefile
```diff
@@ -0,0 +1,12 @@
+.PHONY: check
+
+install:
+	/usr/bin/install -d $(DESTDIR)/usr/share/documentdb/scripts/
+	/usr/bin/install -m 644 utils.sh $(DESTDIR)/usr/share/documentdb/scripts/utils.sh
+	/usr/bin/install -m 644 start_oss_server.sh $(DESTDIR)/usr/share/documentdb/scripts/start_oss_server.sh
+	/usr/bin/install -m 644 build_and_start_gateway.sh $(DESTDIR)/usr/share/documentdb/scripts/build_and_start_gateway.sh
+	/usr/bin/install -m 644 setup_psqlrc.sh $(DESTDIR)/usr/share/documentdb/scripts/setup_psqlrc.sh
+#	/usr/bin/install -m 644 documentdb-setup.sh $(DESTDIR)/usr/share/documentdb/scripts/documentdb-setup.sh
```
AndrewKhoma: drop it?

AndrewKhoma (approved):

scripts/Makefile
sthapa: I'll remove the comment. The issue originally mentioned this but I see that writing the script is another issue entirely ( #562   ). I left this in for future proofing but tbh, it can be added as part of that issue.

AndrewKhoma: Make sense! Can you then not remove it, but also place the comment pointing to the context (this PR), so in future we can reference and move it further?

packaging/rpm/spec/documentdb.spec
```diff
@@ -98,6 +99,22 @@ rm -rf %{buildroot}/usr/src/documentdb/build
...
+# scripts needed for docker container and cli
+%{_datadir}/documentdb/scripts/utils.sh
+%{_datadir}/documentdb/scripts/start_oss_server.sh
+%{_datadir}/documentdb/scripts/build_and_start_gateway.sh
+%{_datadir}/documentdb/scripts/emulator_entrypoint.sh
+%{_datadir}/documentdb/scripts/init_documentdb_data.sh
+%{_datadir}/documentdb/scripts/setup_psqlrc.sh
+# %{_datadir}/documentdb/scripts/documentdb-setup.sh
+# sample data needed for docker container and cli
+%{_datadir}/documentdb/sample-data/01-users.js
```
safern, line 111: Hello, thanks for the contribution. Would you mind expanding why do we need the sample data on the main rpm documentdb.spec?  I think the sample data makes sense to the documentdb-local container, but the rpm package I think shouldn't need it. 

In fact I would argue documentdb-local needs it but that is a separate conversation.

AndrewKhoma (approved):

scripts/Makefile
sthapa: Sure, I'll push changes for this later today

sthapa: Actually on second thought, it's probably best to leave it out for now. Adding script for #562 will need to modify the spec file anyway so this adds extra stuff to PR without saving any future work.

packaging/rpm/spec/documentdb.spec
sthapa: Sure, so my understanding is that we'd like the scripts and sample data to be packaged in the rpm or debs so that they can be used in the Dockerfiles without copying it in from the repo ( #564 ).

sthapa:
What's the best to get the sign-off check to pass here? It doesn't look like either of the options for fixing it is appropriate.

sthapa:
I opened another [PR](https://github.com/documentdb/documentdb/pull/661/) for this in order to fix the missing DCO checks.

PR #654, opened 2026-07-12
Fix explain executionStats InternalError on PostgreSQL 18

### LINKED ISSUE #653: explain with verbosity: executionStats returns InternalError on PostgreSQL 18 with pg_documentdb_gateway
**Describe the bug**

Running an explain command with verbosity: "executionStats" (or allPlansExecution / allShardsExecution) against a PostgreSQL 18 backend always fails with a generic InternalError. The same query with verbosity: "queryPlanner" works fine, and the same executionStats explain works on PostgreSQL ≤ 17.
 
Root cause: starting with PostgreSQL 18, EXPLAIN (ANALYZE, FORMAT JSON) prints per-node runtime counters as fractional numbers (e.g. "Actual Rows": 2.33, and even 3.00 for loops = 1 nodes). The gateway deserializes these counters into Option<i64> fields, so any decimal value fails to parse. With serde_json's arbitrary_precision feature this produces an opaque invalid number error, which is surfaced to the client as InternalError. queryPlanner is unaffected because it does not run ANALYZE and thus has no actual-row counters.

The gateway log shows:
operation_name=Explain
error_message_internal=Failed to parse backend explain plan: invalid number

**Environment**
- Operating system and version: Linux (reproduced on TencentOS / CentOS-class glibc); OS-independent
- PostgreSQL version: 18.4 (fails). PostgreSQL 17.x and earlier are not affected.
- Architecture: x86_64 (architecture-independent)
- DocumentDB extension: 0.115-0 (gateway MongoDB wire compat 7.0.0)

**Reproduction Steps**

Against a gateway backed by PostgreSQL 18:
 
mongosh "mongodb://<user>:<pwd>@<host>:<port>/test_repro?authMechanism=SCRAM-SHA-256&directConnection=true" --quiet
const c = `repro_${Date.now()}`;
db.getCollection(c).insertMany([{ k: 1 }, { k: 2 }, { k: 3 }]);
 
// Works:
db.runCommand({ explain: { find: c, filter: { k: { $gte: 1 } } }, verbosity: 'queryPlanner' });
  
// Fails with InternalError on PG18:
db.runCommand({ explain: { find: c, filter: { k: { $gte: 1 } } }, verbosity: 'executionStats' });
  
The underlying trigger can be confirmed directly in psql on the PG18 backend — note the decimal row counts:
EXPLAIN (ANALYZE, FORMAT JSON) SELECT * FROM o JOIN inr ON inr.k = o.i;
-- "Node Type": "Index Only Scan", "Actual Rows": 2.33, "Actual Loops": 3
-- "Node Type": "Seq Scan",        "Actual Rows": 3.00, "Actual Loops": 1

**Expected behavior**

executionStats explain returns the execution statistics document (ok: 1, executionStats.nReturned, totalDocsExamined, etc.), just as it does on PostgreSQL ≤ 17.

**Actual behavior**
 { "name": "MongoServerError", "code": 1, "codeName": "InternalError", "message": "An unexpected internal error has occurred." }

**Additional context**

starkm-li:
PG18 prints fractional per-loop row counts in EXPLAIN ANALYZE (e.g. "Actual Rows": 2.33), which failed to deserialize into the gateway's Option<i64> counters, surfacing as InternalError. Parse these fields via a float-tolerant deserializer that rounds back to i64, and add unit + functional coverage for the executionStats path.

Fixes #653

### DIFF
--- a/pg_documentdb_gw/documentdb_gateway_core/src/explain/model.rs
+++ b/pg_documentdb_gw/documentdb_gateway_core/src/explain/model.rs
@@ -6,7 +6,24 @@
  *-------------------------------------------------------------------------
  */
 
-use serde::Deserialize;
+use serde::{Deserialize, Deserializer};
+
+/// Deserialize an optional integer counter, tolerating fractional values.
+///
+/// `PostgreSQL` 18 prints per-node runtime counters averaged over `Actual
+/// Loops`, so they can arrive as floats (e.g. `2.33`) when loops > 1. We round
+/// back to `i64` so both the PG18 and pre-18 wire formats parse.
+#[expect(
+    clippy::cast_possible_truncation,
+    reason = "runtime counters are small; rounding an averaged f64 back to i64 is intentional"
+)]
+fn de_opt_round_i64<'de, D>(deserializer: D) -> Result<Option<i64>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let value: Option<f64> = Option::deserialize(deserializer)?;
+    Ok(value.map(|v| v.round() as i64))
+}
 
 #[derive(Deserialize, Debug, Clone)]
 #[serde(rename_all = "PascalCase")]
@@ -66,7 +83,7 @@ pub struct PostgresExplain {
 #[derive(Deserialize, Debug, Clone, Default)]
 #[serde(rename_all = "PascalCase")]
 pub struct ExplainPlan {
-    #[serde(rename = "Actual Rows")]
+    #[serde(rename = "Actual Rows", default, deserialize_with = "de_opt_round_i64")]
     pub actual_rows: Option<i64>,
 
     #[serde(rename = "Actual Total Time")]
@@ -89,7 +106,7 @@ pub struct ExplainPlan {
     #[serde(rename = "Group Key")]
     pub group_key: Option<Vec<String>>,
 
-    #[serde(rename = "Heap Fetches")]
+    #[serde(rename = "Heap Fetches", default, deserialize_with = "de_opt_round_i64")]
     pub heap_fetches: Option<i64>,
 
     #[serde(rename = "Index Cond")]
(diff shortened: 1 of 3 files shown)


visridha (commented):
LGTM. Will wait for @AndrewKhoma to do a pass

pg_documentdb_gw/documentdb_gateway_core/src/explain/model.rs
```diff
@@ -89,7 +106,7 @@ pub struct ExplainPlan {
     #[serde(rename = "Group Key")]
     pub group_key: Option<Vec<String>>,
 
-    #[serde(rename = "Heap Fetches")]
+    #[serde(rename = "Heap Fetches", default, deserialize_with = "de_opt_round_i64")]
```
AndrewKhoma, line 109: so I couldn't find any of the pg commits that modified this field (or any others than Actuals rows), here's the pg commits I saw that did the change:
1. https://git.postgresql.org/gitweb/?p=postgresql.git;a=commitdiff;h=95dbd827f
2. https://git.postgresql.org/gitweb/?p=postgresql.git;a=commitdiff;h=ddb17e387

but I don't see a huge regression of rounding the floats, so all good

pg_documentdb_gw/documentdb_tests/src/commands/explain.rs
```diff
@@ -44,5 +44,27 @@ pub async fn validate_explain(db: &Database) -> Result<(), Error> {
     })
     .await?;
 
+    let lookup_from = db.collection("test_lookup_from");
+    for i in 0..20 {
+        lookup_from.insert_one(doc! {"k": i % 3, "v": i}).await?;
+    }
+
+    db.run_command(doc! {
```
AndrewKhoma, line 97: should we also validate that `totalDocsExamined` is returned correctly?

AndrewKhoma (approved):

pg_documentdb_gw/documentdb_gateway_core/src/explain/model.rs
starkm-li: Yes, the root change is in this commit.
https://git.postgresql.org/gitweb/?p=postgresql.git;a=commitdiff;h=95dbd827f

PG17 printed the value using %.0f (rounding to the nearest integer), while PG18 changed this to %.2f to consistently display two decimal places.

pg_documentdb_gw/documentdb_tests/src/commands/explain.rs
starkm-li: Done — the test now asserts the counters instead of just success: nReturned == 3 and totalDocsExamined >= nReturned (lower bound to stay stable across PG versions). On PG18 these come back as 3.00, so it verifies the fix rounds them to correct integers.

AndrewKhoma (approved):

visridha (approved):

safern (approved):

visridha:
Looks like you need to do some cargo lint to make the tests pass. @starkm-li would you be able to clean that up and repush please?

tsharp (approved):

xgerman (approved):

tsharp (approved):

AndrewKhoma (approved):

xgerman (approved):

AndrewKhoma (approved):

tsharp (approved):

PR #663, opened 2026-07-15
Fix $expr with a scalar argument failing the query when an index exists

gmprestes:
## Problem

A `$expr` filter whose argument is a scalar — all valid MongoDB — fails the entire query whenever the collection has an eligible composite index:

```js
db.c.insertOne({v: 1})
db.c.createIndex({v: 1})
db.c.find({$expr: "$v"})    // (BadValue) expected a document or array to init iterator, got string
db.c.find({$expr: -1})      // same, got double
db.c.find({$expr: true})    // same, got bool
```

Without the index the same queries work, because the planner support function never runs — which makes the failure look nondeterministic from the client's perspective.

## Root cause

`PushExprToIndex` assumes the `$expr` argument is always a document and calls `BsonValueInitIterator` on it directly; that helper throws for any non-document/array input, and the error escapes query planning.

## Fix

A scalar `$expr` argument carries no operator conditions that could be pushed to the index: return `NULL` (no index quals) and let the runtime evaluate the expression, exactly as it does when no index exists.

## Validation

- This accounted for 312 failing cases in the FerretDB compatibility suite (`TestQueryEvaluationCompatExpr`, every scalar type), compared side-by-side against a real MongoDB 7 replica set; all pass with this change.
- Full suite (Query/Update/Aggregate/Insert/Delete/Count/CreateIndexes/DropIndexes) runs clean with the fix applied, on two environments (macOS/Docker Desktop and a 40-core Linux host).


pg_documentdb/src/opclass/index_operator_support.c
```diff
@@ -415,6 +415,16 @@ PushExprToIndex(SupportRequestIndexCondition *supportRequest)
 		return NULL;
 	}
 
+	/* $expr accepts scalar expressions (constants, field paths); only a
```
visridha, line 418: Could you please add some backend tests for this in the documentdb test suite covering some failed cases and asserting the expected behavior?

visridha (changes requested):

gmprestes:
Added in `b9473c10`. The existing coverage was incomplete — no test exercised a non-document `$expr` argument at all — so this fills in the missing cases: a field path, every scalar constant type, and an array. Each asserts the plan (no index quals pushed down, expression left to the runtime filter) and the result.

`{ "a": 5, "$expr": "$a" }` is the one worth a look: the sibling equality still reaches the `a_1` index, so the guard costs no pushdown that was previously possible.

visridha (approved):

safern (approved):

PR #664, opened 2026-07-15
Fix Infinity handled as a valid integer in $type, $size and bitwise operators

gmprestes:
## Problem

`IsDoubleAFixedInteger` checks `floor(value) == value`, which Infinity satisfies. The value then flows into `BsonValueAsInt64` and saturates to `INT64_MAX`, producing for `$type`:

```
(BadValue) Invalid numerical type code: 9223372036854775807
```

where MongoDB reports:

```
(BadValue) Invalid numerical type code: inf
```

For `$size` and the bitwise query operators, an Infinity argument was silently accepted as a fixed integer instead of failing to parse the way MongoDB does.

## Fix

- `IsDoubleAFixedInteger` now requires a finite value, so `$size` and bitwise operators reject Infinity like MongoDB;
- `$type` reports non-finite numeric codes with MongoDB's exact spelling (`inf`/`-inf`/`nan`) before the fixed-integer check.

## Validation

FerretDB compatibility suite compared side-by-side against a real MongoDB 7 replica set: the `$type` Infinity cases (`TestQueryElementCompatElementType`) and `TestQueryArrayCompatSize/Infinity` now behave identically to MongoDB. Full suite (8 groups) runs clean with the change, on two environments.


gmprestes:
Added in `c9052a01`, filling gaps the existing tests left open.

A numeric `$type` code turns out to be testable after all. The comment in `bson_query_operator_error_tests.sql` is right that a bare number fails to parse, but `{ "$type": { "$numberDouble": "Infinity" } }` parses and reaches the numeric path — so `inf`, `-inf` and `nan` are now asserted directly instead of a saturated `9223372036854775807`, with `2.0`, `2.5` and `9999` as controls.

`$size` and the bitwise double branch are covered too; the decimal128 spelling was already rejected and already covered.

visridha (approved):

safern (approved):

PR #666, opened 2026-07-15
Allow parallel plans on the single-batch and file-based cursor drain paths

gmprestes:
## Problem

Neither `CreateAndDrainSingleBatchQuery` nor `PlanForcedPersistentQuery` passes `CURSOR_OPT_PARALLEL_OK`, so `pg_plan_query` never produces a parallel plan for any aggregation: a full-collection `$group` runs its HashAggregate — group-key expression evaluation, bson decompression and decoding per document — in a single process regardless of `max_parallel_workers_per_gather`. Misleadingly, EXPLAIN shows a parallel plan, because the explain path plans with different cursor options than the execution path.

## Fix

Both paths drain the entire result in one executor run (`DrainStatementViaExecutor`: `execute_once=true`, `count=0`): no portal is suspended and no cursor scrolls backwards — precisely the execution shape a parallel plan requires.

- `CreateAndDrainSingleBatchQuery`: allow parallel plans for read-only queries.
- `PlanForcedPersistentQuery`: allow them only when the plan is destined for the file-based cursor path (hold cursor + `useFileBasedPersistedCursors`), where continuation pages go to a file instead of a suspended portal. If such a plan ever lands on the suspended-portal path, the executor detects the bounded fetch and runs the Gather leader-only — the failure mode is "not parallel", never "wrong results".

## Measured impact

98k-document collection, `$group` by a bson field path, warm cache, 4 workers, group-count statistics on the key: **690ms → 128ms**, at parity with MongoDB 7 on the same machine (109ms).

## Validation

Full FerretDB compatibility suite (8 groups, side-by-side against MongoDB 7) with `useFileBasedPersistedCursors` enabled: zero failures, on two environments (macOS/Docker Desktop and a 40-core Linux host).


visridha (approved):

safern (approved):

safern:
Thanks for your contribution @gmprestes

PR #668, opened 2026-07-15
Fix $mod returning different results on index scan vs runtime for -2^63

gmprestes:
## Problem

The same `$mod` query returns different documents depending on whether the planner picks an index scan or a sequential scan:

```js
// docs
{_id: "int64-min", v: NumberLong("-9223372036854775808")}
{_id: "double-1",  v: -9223372036854775808.0}   // exactly -2^63, representable in both types

// query
{v: {$mod: [NumberLong("-9223372036854775808"), 0]}}

// index scan:  both documents  (matches MongoDB)
// seq scan:    only int64-min
```

## Root cause

`CompareModOperator` validates the document value with `IsBsonValue64BitInteger`, which quantizes doubles to 15 significant digits before the int64 range check. `-9223372036854775808.0` quantizes to `-9.22337203685478e18`, overflows the check, and the runtime declares it "not a 64-bit integer" — while the index path, whose terms are already normalized to int64, matches it.

## Fix

Use the unquantized range check (`IsBsonValueUnquantized64BitInteger`). MongoDB truncates the dividend toward zero to a 64-bit integer without rounding; the exact boundary value -2^63 is representable and must match.

## Validation

Fixes the runtime side of FerretDB/FerretDB-DocumentDB#331: the two xfail cases for that issue in the FerretDB compatibility suite now pass against a real MongoDB 7. Full suite (8 groups) clean on two environments.


visridha:
Could you add a backend test please?

gmprestes:
Added in `e87b4440`. The boundary fixtures were incomplete: no document held the exact -2^63 double (`_id 135` is `-922337203685477e4`). It goes in as `_id 138`, so every `$mod` query in the shared `_core` body now evaluates against it and the runtime and index drivers — generated from that one file — have to agree.

Against the unfixed code, `_id 138` matches in 14 result sets on the index path and none on the runtime path, which is the divergence this PR removes; with the fix both report 14. `_id 139` holds 2^63 and must never match.

visridha (approved):

safern (approved):

Issue #697, opened 2026-07-30
[Feature Request] Configurable MongoDB server version and max wire version


Asamsig:
**Purpose of the feature.**

Currently MongoDB server version and max wire version are effectively hardcoded to version 7.0.0 / 21 ([dead config variable](https://github.com/documentdb/documentdb/blob/fd46318bf292780238eac4ef2f9e0011f7234539/pg_documentdb_gw/documentdb_gateway_core/src/configuration/dynamic.rs#L149-L154), that lacks GUC-backing), this can cause issues since some clients do assertions/checks to verify compatibility, effectively rejecting the current 'version'.
You should be able to configure these values, to satisfy clients, even if there is no 'official' support for the given version by DocumentDB.

**Describe the solution you'd like**

Two new GUCs the gateway reads via dynamic config:
- `documentdb.server_version`: string. If empty/unset, default to `7.0.0` (this is the current default). I don't think DocumentDB needs to enforce a format for the version string, it could be left up to the users control.
- `documentdb.max_wire_version`: int. If unset, default to `21` (this is the current default).

I suggest dropping the [version-enum](https://github.com/documentdb/documentdb/blob/fd46318bf292780238eac4ef2f9e0011f7234539/pg_documentdb_gw/documentdb_gateway_core/src/configuration/version.rs) completely, and have these values be independent of each other.

**Describe alternatives you've considered**

Keeping a single version enum/allowlist in the gateway that ties Mongo version and wire version together. That is exists today, it is just missing a GUC to actually be able to change it, but it is inflexible, you can't set a patch version or bump the wire version without a code change.
Put this only in gateway config or env vars. Possible, but most DocumentDB knobs already live as GUCs in pg_settings, so GUCs feel like the consistent place.

**Additional context**

Currently we have [version.rs](https://github.com/documentdb/documentdb/blob/fd46318bf292780238eac4ef2f9e0011f7234539/pg_documentdb_gw/documentdb_gateway_core/src/configuration/version.rs), which only allows a few values, and ties version and wire version together. I don't see the harm in letting end-users configure whatever they want, to me this will create an escape hatch, where you're interacting with a strict client.

PR #710, opened 2026-08-02
Add documentdb.server_version and documentdb.max_wire_version GUCs for gateway-reported versions

Asamsig:
## Description

Add `documentdb.server_version` and `documentdb.max_wire_version`  GUCs so you can override what the gateway reports as MongoDB serverVersion and maxWireVersion.

- `documentdb.server_version`, current default `7.0.0`, as today, reported in `buildInfo`.
- `documentdb.max_wire_version`, current default `21`, as today, reported in `hello` and `isMaster`.

The defaults live in the extension, added fallbacks in the gateway as well to follow the established pattern of other GUCs.

`server_version` is validated as `major.minor[.patch[.build]]`, this is because the gateway parses the version to build `versionArray` failing to parse it, would create a mismatch between `version` and `versionArray`, enforcing the pattern in the GUC avoids that possibility.

Removed the Version-enum in the gateway, to make version more flexible, also decoupled `server_version` from `max_wire_version`.

To change the reported versions:
```sql
ALTER SYSTEM SET documentdb.server_version = '8.0.4';
ALTER SYSTEM SET documentdb.max_wire_version = 25;
SELECT pg_reload_conf();
-- wait for gateway refresh
```

Check the values with:
```
db.runCommand({ hello: 1 }).maxWireVersion

db.runCommand({ buildInfo: 1 }).version

db.runCommand({ buildInfo: 1 }).versionArray
```

Related: #697

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] Documentation update
- [ ] Refactoring / code cleanup
- [ ] DevOps / tooling
- [ ] Test improvement
- [ ] Other: ___________

## AI Disclosure

- [ ] No AI tools were used substantially in this PR
- [X] AI tools were used to assist with this PR (please complete below)
  - **Tool(s) used:**  Cursor
  - **How it was used:**  Used for assistance with GUC design, and getting familiar with the project / workflows.

## Checklist

- [X] I can explain every change in this PR if asked during review
- [X] I have tested these changes locally
- [X] I have added or updated tests as appropriate
- [X] All commits are signed off (DCO: `git commit -s`)
- [X] I have read the [CONTRIBUTING.md](../CONTRIBUTING.md) guidelines

### DIFF
--- a/pg_documentdb/src/test/regress/validate_test_output.sh
+++ b/pg_documentdb/src/test/regress/validate_test_output.sh
@@ -13,7 +13,7 @@ check_directory=$2
 aggregateCollectionIdStr=""
 maxCollectionIdStr=""
 
-validationExceptions="/sql/documentdb_test_helpers.sql /sql/public_api_schema.sql /sql/documentdb_guc_naming_conventions.sql /sql/documentdb_rum_guc_naming_conventions.sql"
+validationExceptions="/sql/documentdb_test_helpers.sql /sql/public_api_schema.sql /sql/documentdb_guc_naming_conventions.sql /sql/documentdb_rum_guc_naming_conventions.sql /sql/documentdb_server_version_guc.sql"
 
 echo "Validating test file output"
 for validationFile in $(ls $check_directory/expected/*.out); do
--- a/pg_documentdb/src/configs/system_configs.c
+++ b/pg_documentdb/src/configs/system_configs.c
@@ -136,6 +136,18 @@ bool ThrowDeadlockOnCrud = DEFAULT_THROW_DEADLOCK_ON_CRUD;
 #define DEFAULT_LOCALHOST_CONN_STR "host=localhost"
 char *LocalhostConnectionString = DEFAULT_LOCALHOST_CONN_STR;
 
+/*
+ * MongoDB server version reported by the gateway in buildInfo.
+ */
+#define DEFAULT_SERVER_VERSION "7.0.0"
+char *ServerVersion = DEFAULT_SERVER_VERSION;
+
+/*
+ * maxWireVersion reported by the gateway in hello / isMaster.
+ */
+#define DEFAULT_MAX_WIRE_VERSION 21
+int MaxWireVersion = DEFAULT_MAX_WIRE_VERSION;
+
 /* Currently timeout max at 3 hours */
 #define DEFAULT_MAX_CUSTOM_COMMAND_TIMEOUT (3600 * 3 * 1000)
 int MaxCustomCommandTimeout = DEFAULT_MAX_CUSTOM_COMMAND_TIMEOUT;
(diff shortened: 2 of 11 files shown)


AndrewKhoma (changes requested):
thank you for your changes, but they are not compatible with the mongodb drivers, see:
https://github.com/mongodb/mongo-csharp-driver/blob/7dd57f03993e51cec33c5a23317183604ef0c639/src/MongoDB.Driver/Core/Misc/WireVersion.cs#L138-L139

we can't just set 8.0.3 as you set in tests and report it back to the driver since driver does not know that version and all the WireVersions are predefined for the specified version.

we never updated it since we have other versioning system as mongodb, but if you'll add the static mapping between those 2 (they can't be set independently, i.e. there's a clear bijection between wire version 27 and mongo version 8.2), then it will be a great change!

Asamsig:
> thank you for your changes, but they are not compatible with the mongodb drivers, see: https://github.com/mongodb/mongo-csharp-driver/blob/7dd57f03993e51cec33c5a23317183604ef0c639/src/MongoDB.Driver/Core/Misc/WireVersion.cs#L138-L139
> 
> we can't just set 8.0.3 as you set in tests and report it back to the driver since driver does not know that version and all the WireVersions are predefined for the specified version.
> 
> we never updated it since we have other versioning system as mongodb, but if you'll add the static mapping between those 2 (they can't be set independently, i.e. there's a clear bijection between wire version 27 and mongo version 8.2), then it will be a great change!

I understand that wire version and server version are related, but it doesn't seem like drivers actually require them to match, and my pitch is that if this causes problems it would be a user error. (I'm open to being wrong. :) )

If we take the C# driver as an example the server version is actually derived from the wire version:

https://github.com/mongodb/mongo-csharp-driver/blob/8b699a54fe741ab819a2bb0739478ec5cb551703/src/MongoDB.Driver/Core/Servers/ServerMonitor.cs#L418-L419

It is also the wire version that drives feature gates. So as far as I can see, this wouldn't create problems with the driver.

`node-mongodb-native` is similar: https://github.com/mongodb/node-mongodb-native/blob/main/src/cmap/connect.ts#L65-L88, in how it reads wire version and maps that to server version itself, it doesn't read the reported server version in `buildInfo` for anything.

---

So if we were to change to a single value it would make most sense to just set `wire_version`, and then documentdb would maintain a mapping like there was, just the other way around, than how it was done before, so from wire version to server version, instead of from server version to wire version.

Tempting to support setting either wire or server version, and infer the one from the other, but seems unnecessarily complex to support both.

My inclination is that users will experience drivers complaining about a too low wire version, and based on the error, they'll see what wire version they'd need to update to, perhaps with the server version in parenthesis. So wire version is probably the value that would be most familiar.

Would to want some kind of validation on the GUC, or only do validation in the gateway?

AndrewKhoma:
> So if we were to change to a single value it would make most sense to just set wire_version, and then documentdb would maintain a mapping like there was, just the other way around, than how it was done before, so from wire version to server version, instead of from server version to wire version.

yes, let's just keep the wire version.

> Would to want some kind of validation on the GUC, or only do validation in the gateway?

gateway validation will be sufficient, since drivers are speaking through gateway and not via backend. in the backend I would just make sure we have a correct ranges of where the wire version can go

Asamsig:
@AndrewKhoma Finally got around to taking another stab at this.

I've now amended the commit, and instead of two separate GUCs, we now have just one GUC for setting the `max_wire_version`, the `server_version` is then paired directly to the `max_wire_version`.

pg_documentdb/src/test/regress/validate_test_output.sh
```diff
@@ -13,7 +13,7 @@ check_directory=$2
 aggregateCollectionIdStr=""
 maxCollectionIdStr=""
 
-validationExceptions="/sql/documentdb_test_helpers.sql /sql/public_api_schema.sql /sql/documentdb_guc_naming_conventions.sql /sql/documentdb_rum_guc_naming_conventions.sql /sql/documentdb_stat_bgworker_jobs.sql"
+validationExceptions="/sql/documentdb_test_helpers.sql /sql/public_api_schema.sql /sql/documentdb_guc_naming_conventions.sql /sql/documentdb_rum_guc_naming_conventions.sql /sql/documentdb_stat_bgworker_jobs.sql /sql/documentdb_max_wire_version_guc.sql"
```
AndrewKhoma: why is it part of the validation exceptions?

Asamsig: It only tests the GUC, but there is not really any special logic to it anymore, so it could probably be dropped completely.
The test never creates a collection, so it fails with `Test file prefix Validation failed on 'documentdb_max_wire_version_guc.sql': Please ensure test files set documentdb.next_collection_id
`.

Considering what is left over in this test, my inclination is to just delete the test, seems silly to just test the bounds of an integer GUC.

Asamsig: I dropped the test, since it was merely testing the bounds of the GUC.

PR #721, opened 2026-08-13
fix(gateway): reuse existing user pool on settings-key miss in get_data_pool

### LINKED ISSUE #720: Gateway permanently fails all requests for a user after a dynamic-config refresh flips the connection-pool key ("Connection pool missing for user.")
## Summary

After `pg_documentdb_gw`'s periodic dynamic-configuration refresh runs (default interval `60*5 = 300s`, see `configuration/setup.rs`), every subsequent request from an already-authenticated data-plane user can start failing with the internal error `"Connection pool missing for user."` — and it never recovers on its own. The user's connection pool is still present in `user_data_pools` and perfectly healthy; it's just filed under a cache key that the per-request lookup will never recompute back to. The client only sees a generic server-origin error (the gateway returns an error document with `ok: 0`), so this is not a client-side bug.

## Environment

- Image: `documentdb-local v0.114-0` (built 2026-07-17), the latest release at the time of writing.
- `DynamicConfigurationRefreshIntervalSecs` left at the built-in default (`60 * 5` seconds — `SetupConfiguration::dynamic_configuration_refresh_interval_secs`, `configuration/setup.rs`).
- Client: a long-lived Mongo-wire client, authenticating as a fixed data-plane user and issuing periodic authenticated requests.
- Observed over ~17h on a long-running deployment: the failure recurs phase-locked to the 300s refresh timer — roughly 4m35s of healthy service, then failing until the client reconnects / the process restarts.

## Root cause

`PgPoolSettings` (`postgres/conn_mgmt/pool_settings.rs`) derives `Hash`/`Eq` over six fields:

```rust
#[derive(Copy, Clone, Debug, Eq, Hash, PartialEq)]
pub struct PgPoolSettings {
    max_connections: usize,
    system_connection_budget: usize,
    connection_buffer_size: usize,
    connection_pruning_interval: Duration,
    connection_idle_lifetime: Duration,
    connection_lifetime: Duration,
}
```

`from_configuration()` builds this struct fresh from the *live* dynamic configuration on every call:

```rust
pub fn from_configuration(config: &dyn DynamicConfiguration) -> Self {
    let max_connections = config.max_connections();
    ...
}
```

`max_connections` is read from a live Postgres GUC every dynamic-config refresh, via `PgConfiguration::max_connections()` (`configuration/pg_configuration.rs`):

```rust
fn max_connections(&self) -> usize {
    let max_connections = self.get_i32("max_connections", -1);
    match max_connections {
        n if n < 0 => {
            // theoretically we can't end up here, since Postgres always provide values
            tracing::error!("GUC max_connections is not setup correctly");
            25usize
        }
        n => usize::try_from(n).unwrap_or(25),
    }
}
```

On a bad/transient read this silently substitutes `25`. (The same code path also fires on any *legitimate* runtime change to the `max_connections` GUC, e.g. an operator running `ALTER SYSTEM SET max_connections = ...; SELECT pg_reload_conf();` — that's an intentional, tested behavior; see `validate_max_conn_change` in `pool_manager.rs`'s test module.) Either way, once this value changes, `PgPoolSettings`'s hash changes too.

The refresh itself runs on `PgConfiguration`'s background timer (`start_dynamic_configuration_refresh_thread`, `configuration/pg_configuration.rs`), on the interval configured by `dynamic_configuration_refresh_interval_secs` (default `60*5`, `configuration/setup.rs`), calling `refresh_configuration()` which swaps in the new value map and reloads the topology snapshot.

`PoolManager::get_data_pool` (`postgres/conn_mgmt/pool_manager.rs`) recomputes `(username, PgPoolSettings)` on *every* request and does an exact-key lookup:

```rust
pub fn get_data_pool(
    &self,
    username: &str,
    dynamic_configuration: &dyn DynamicConfiguration,
) -> Result<Arc<ConnectionPool>> {
    let settings = PgPoolSettings::from_configuration(dynamic_configuration);

    match self.user_data_pools.get(&(username.to_owned(), settings)) {
        None => Err(DocumentDBError::internal_error(
            "Connection pool missing for user.".to_owned(),
        )),
        Some(pool_ref) => {
            let pool = Arc::clone(pool_ref.value());
            pool.touch();
            Ok(pool)
        }
    }
}
```

Once `settings` no longer matches the key the pool was inserted under, this misses — permanently, since nothing ever re-inserts under the *old* key, and nothing recomputes the *new* key's entry either, because `allocate_data_pool` (the only function with the user's password, needed to build a new pool) only runs during authentication, not on this per-request path. Every request for that user fails from that point on, until the client reconnects (forcing re-authentication) or the gateway process restarts.

For contrast, `get_system_shared_pool` in the same file *does* self-heal on a settings-key miss:

```rust
match self.shared_data_pools.entry(settings) {
    Entry::Occupied(pool_ref) => { ... }
    Entry::Vacant(entry) => {
        let system_shared_pool = Arc::new(ConnectionPool::new_with_user(
            self.setup_configuration.as_ref(),
            &self.query_catalog,
            self.setup_configuration.postgres_data_user(),
            self.setup_configuration.postgres_data_user_password(),
            ...
        )?);
        entry.insert(Arc::clone(&system_shared_pool));
        Ok(system_shared_pool)
    }
}
```

This works because the shared pool uses system-known credentials (`setup_configuration.postgres_data_user()` / `postgres_data_user_password()`), which are always available. `get_data_pool`'s user path has no equivalent: the actual user password is never held outside of the `allocate_data_pool` authentication call, so there is nothing it could use to build a replacement pool inline.

## Steps to reproduce

1. Run `pg_documentdb_gw` (`documentdb-local v0.114-0`) against a Postgres backend with the default `DynamicConfigurationRefreshIntervalSecs` (300s).
2. Authenticate a client as a data-plane user/password and keep issuing periodic authenticated requests (as a long-lived service would).
3. While the client is connected, cause the `max_connections` GUC value observed by the gateway's next refresh to change — either genuinely (`ALTER SYSTEM SET max_connections = <new>; SELECT pg_reload_conf();`) or via a transient failure of the `pg_settings` query that the refresh issues (see `PgConfigurationInner::load_configurations`), which triggers the silent-25 fallback in `max_connections()`.
4. Wait for the next dynamic-config refresh tick (up to 300s by default).
5. Issue another authenticated request as the same user.

## Expected behavior

The request succeeds. The user's existing connection pool is still healthy; a change in `max_connections` (real or an artifact of a bad GUC read) should not invalidate the client's ability to keep using it.

## Actual behavior

The request — and every subsequent one from that user — fails immediately with the internal error `"Connection pool missing for user."`, surfaced to the client as a generic error document (`ok: 0`) originating from the server, which rules out a client-side cause. The failure recurs phase-locked to the 300s refresh interval — roughly 4m35s healthy, then failing — and only a client reconnect (re-authentication) or a process restart clears it.

## Log evidence

The gateway logs the topology/config refresh via `tracing::info!("Topology acquired: {doc:?}")` in `PgConfiguration::load_topology` (`configuration/pg_configuration.rs`), which runs as part of every `refresh_configuration()` call. This line reliably precedes the first `"Connection pool missing for user."` error by roughly 10 seconds — consistent with the refresh completing, the pool key shifting, and the next incoming request from the still-connected client hitting the stale-key miss shortly after (representative excerpt):

```
...  INFO documentdb_gateway_core::configuration::pg_configuration: Topology acquired: PgDocument { ... }
...  (≈10s later)
...  ERROR documentdb_gateway_core::...: Connection pool missing for user.
```

## Possible duplicate

This may overlap with #379 ("[Feature Request] Automatically retry writes on the gateway"), which is about gateway resilience when the backing Postgres is unavailable during failover. That issue's failure mode (PG actually down) is different from this one (PG is up and the user's pool is healthy — only an internal cache key is stale), but both concern the gateway hard-failing user requests instead of tolerating a transient/self-correcting backend condition. Flagging for maintainers to check for duplication or overlap in the underlying fix.

## Additional context

This investigation (log correlation, root-cause tracing through the config-refresh → pool-key → pool-manager path, and the reproduction steps above) was prepared with substantial AI assistance and reviewed by the author before submission.

morch23mj:
## Description

Fixes #720.

`get_data_pool` (`postgres/conn_mgmt/pool_manager.rs`) keys the per-user pool cache on `(username, PgPoolSettings)`. `PgPoolSettings` includes `max_connections`, which is re-read from a Postgres GUC on every dynamic-config refresh (and silently defaults to `25` on a bad read). When it changes, the exact-key lookup misses and the gateway hard-fails **every** request for that user — permanently — even though their pool is still present and healthy. This PR makes `get_data_pool` reuse the user's existing pool on a settings-key miss instead of erroring. Root cause and evidence: #720.

## What changed

`get_data_pool` keeps the exact-key fast path, and on a miss reuses any existing pool for that `username` before erroring (illustrative):

```rust
if let Some(p) = self.user_data_pools.get(&(username.to_owned(), settings)) {
    p.touch(); return Ok(Arc::clone(p.value()));
}
// settings drifted (e.g. max_connections re-read) — pool is healthy but filed
// under a stale key. Reuse it; allocate_data_pool re-files under the current
// key on the next auth.
if let Some(p) = self.user_data_pools.iter()
    .find(|e| e.key().0 == username).map(|e| Arc::clone(e.value())) {
    p.touch(); return Ok(p);
}
Err(DocumentDBError::internal_error("Connection pool missing for user.".to_owned()))
```

Plus a regression test (`test_get_data_pool_falls_back_to_existing_pool_on_settings_change`): allocate a pool, flip `max_connections`, assert `get_data_pool` returns the *same* pool (`Arc::ptr_eq`) with no new pool created.

## Why it's correct

- The settings delta isn't a real topology change — role/password/live connections are unaffected; only the pool's *sizing* may lag by one refresh.
- Convergence is preserved: `allocate_data_pool` re-files the pool under the current key on the next authentication, so this only bridges the gap for already-authenticated users — it isn't a permanent bypass. The exact-key hit path is untouched, so the deliberate rebuild-on-`max_connections`-change (`validate_max_conn_change`) still works.

## Why not mirror `get_system_shared_pool`

That path self-heals by *building* a new pool from system credentials it always holds. `get_data_pool` has no user password at request time (only `allocate_data_pool` has it, at auth), so reusing the existing pool is the only option here.

## Alternatives considered

- **Drop `max_connections` from the key** — rejected: rebuilding pools on a real `max_connections` change is intentional/tested (`validate_max_conn_change`).
- **Harden `max_connections()` to not fabricate `25`** — good complementary defense-in-depth, but a *legitimate* `max_connections` change flips the same key and hits the same miss, so it's not a substitute. Happy to add as a follow-up.

## Testing

`cargo test -p documentdb_gateway_core pool_manager` → **13 passed, 0 failed** (the new test, the no-pool-still-errors case, and `validate_max_conn_change`). `rustfmt` + `clippy --workspace --all-targets --all-features -- -D warnings` clean on the pinned toolchain.

## Type of Change
- [x] Bug fix

## AI Disclosure
- [x] AI tools were used. **Tool:** Claude (Anthropic). Root-cause tracing, the fix, and the test were prepared with substantial AI assistance, then reviewed and verified against the source by the author.

## Checklist
- [x] I can explain every change if asked
- [x] Tested locally (`cargo test` — 13 passed)
- [x] Tests added/updated
- [x] Commits signed off (DCO)
- [x] Read CONTRIBUTING.md

---

Signed-off-by: Morchid Chellali <mc@wetransform.to>

### DIFF
--- a/pg_documentdb_gw/documentdb_gateway_core/src/postgres/conn_mgmt/pool_manager.rs
+++ b/pg_documentdb_gw/documentdb_gateway_core/src/postgres/conn_mgmt/pool_manager.rs
@@ -145,16 +145,35 @@ impl PoolManager {
     ) -> Result<Arc<ConnectionPool>> {
         let settings = PgPoolSettings::from_configuration(dynamic_configuration);
 
-        match self.user_data_pools.get(&(username.to_owned(), settings)) {
-            None => Err(DocumentDBError::internal_error(
-                "Connection pool missing for user.".to_owned(),
-            )),
-            Some(pool_ref) => {
-                let pool = Arc::clone(pool_ref.value());
-                pool.touch();
-                Ok(pool)
-            }
+        if let Some(pool_ref) = self.user_data_pools.get(&(username.to_owned(), settings)) {
+            let pool = Arc::clone(pool_ref.value());
+            pool.touch();
+            return Ok(pool);
         }
+
+        // The `settings` half of the key is derived from the live configuration and
+        // can shift out from under an already-authenticated connection: notably
+        // `max_connections` is re-read from a Postgres GUC on every dynamic-config
+        // refresh and silently defaults to 25 on a bad read (see
+        // `PgConfiguration::max_connections`). When it changes, the user's pool is
+        // still healthy but filed under a key that will never be recomputed, and a
+        // hard miss here permanently breaks every request for that user until they
+        // reconnect. Fall back to the existing pool for this username instead;
+        // `allocate_data_pool` re-files it under the current key on the next auth.
+        let fallback = self
+            .user_data_pools
+            .iter()
+            .find(|entry| entry.key().0 == username)
+            .map(|entry| Arc::clone(entry.value()));
+
+        if let Some(pool) = fallback {
+            pool.touch();
+            return Ok(pool);
+        }
+
+        Err(DocumentDBError::internal_error(
+            "Connection pool missing for user.".to_owned(),
+        ))
     }
 
     /// # Errors
@@ -678,6 +697,50 @@ mod tests {
         assert_eq!(
             dynamic_configuration.max_conn() * 2,
             user_pool.status().status().max_size
+        );
+    }
+
+    #[tokio::test]
+    async fn test_get_data_pool_falls_back_to_existing_pool_on_settings_change() {
+        // We still need an async context to create the connection pool (see
+        // ConnectionPool::new_with_user), but the test itself doesn't need to be async.
+        yield_now().await;
+
+        let dynamic_configuration = MaxConnectionConfig {
+            max_conn: 100.into(),
+        };
+        let pool_manager = test_pool_manager();
+
+        pool_manager
+            .allocate_data_pool("user", "password", &dynamic_configuration)
+            .unwrap();
+
+        let original_pool = pool_manager
+            .get_data_pool("user", &dynamic_configuration)
+            .unwrap();
+
+        // Simulate a dynamic-config refresh that flips `max_connections` (e.g. a bad
+        // GUC read silently defaulting to 25). This changes the `PgPoolSettings` half
+        // of the pool key so the exact-key lookup misses, but the user's pool is
+        // still healthy and must be reused instead of hard-erroring.
+        dynamic_configuration.set_max_conn(25);
+
+        let fallback_pool = pool_manager
+            .get_data_pool("user", &dynamic_configuration)
+            .expect(
+                "get_data_pool must fall back to the existing user pool on a settings-key miss",
+            );
+
+        assert!(
+            Arc::ptr_eq(&original_pool, &fallback_pool),
+            "settings-key miss must reuse the same warm pool for the user"
+        );
+
+        // No new pool was created by the fallback path.
+        assert_eq!(
+            3,
+            pool_manager.report_pool_stats().len(),
+            "2 system pools + 1 user pool; fallback must not allocate a new pool"
         );
     }
 


pg_documentdb_gw/documentdb_gateway_core/src/postgres/conn_mgmt/pool_manager.rs
```diff
@@ -145,16 +145,35 @@ impl PoolManager {
...
+        // `PgConfiguration::max_connections`). When it changes, the user's pool is
+        // still healthy but filed under a key that will never be recomputed, and a
+        // hard miss here permanently breaks every request for that user until they
+        // reconnect. Fall back to the existing pool for this username instead;
+        // `allocate_data_pool` re-files it under the current key on the next auth.
+        let fallback = self
+            .user_data_pools
+            .iter()
+            .find(|entry| entry.key().0 == username)
+            .map(|entry| Arc::clone(entry.value()));
```
AndrewKhoma, line 167: but the pool you've received will not govern the gw of using the 25 connections as in your test case, but will let to go the whole 100 connections (which it doesn't have).

the whole rationale of why we don't have this logic for the user data pools is because the max_connections is the startup GUC, so I would expect the whole docker restart so the pg can pick up the new param too; if your use case requires the dynamic change I highly recommend changing the 

    user_data_pools: DashMap<ClientKey, Arc<ConnectionPool>> to something like
    user_data_pools: DashMap<(ClientKey, PgPoolSettings), Arc<ConnectionPool>>

so in case any pg pool settings changes you will use the "fresh" pool and the old pool will be cleaned up on retention

AndrewKhoma (changes requested):

morch23mj:
Thanks @AndrewKhoma — agreed on both points. Closing this in favor of #723, which takes that feedback directly — instead of reusing pools it fixes the root cause: the transient bad `max_connections` read that silently substitutes `25` and flips the pool-cache key mid-run with no real change behind it. The intentional hard-fail in `get_data_pool` stays untouched, and there's no pool reuse / over-subscription. A genuine (restart-driven) change still flips the key as designed.

PR #723, opened 2026-08-15
fix(gateway): keep last-known-good max_connections on bad GUC read

### LINKED ISSUE #720: Gateway permanently fails all requests for a user after a dynamic-config refresh flips the connection-pool key ("Connection pool missing for user.")

morch23mj:
## Description

Fixes #720. Supersedes #721 (see "Why this instead of #721").

`max_connections` is a **startup GUC** — it cannot change without a Postgres restart — yet `PgConfiguration::max_connections()` (`configuration/pg_configuration.rs`) re-reads it on every dynamic-config refresh and **silently substitutes `25` on a bad/missing read**. Because `max_connections` is part of the `PgPoolSettings` pool-cache key, that fabricated `25` flips the key mid-run with no real change behind it; `get_data_pool` then permanently hard-fails every request for that user (`"Connection pool missing for user."`) — the pool is healthy but orphaned. Full root cause and log evidence: #720.

This fixes the root cause: on a bad read, `max_connections()` returns the **last successfully-read value** instead of fabricating a divergent one, so the pool key stays stable and the spurious miss never happens.

## Why this instead of #721

#721 made `get_data_pool` fall back to reusing the user's existing pool on a settings-key miss. As @AndrewKhoma noted in review, that is unsafe: if `max_connections` had genuinely dropped to 25, the reused (100-sized) pool would let requests use connections Postgres no longer has (over-subscription); and the user-pool hard-fail is intentional precisely because `max_connections` is a startup GUC. This PR takes that feedback directly — it does **not** reuse pools and leaves the intentional hard-fail guard in `get_data_pool` untouched. It only removes the *spurious* key flip at its source (the bad read). A genuine `max_connections` change (via restart → fresh read) still updates the value and flips the key as designed, so the intended rebuild path is preserved.

## What changed

`max_connections()` caches the last valid read in an `AtomicUsize` (stored on every valid read); on a bad read it returns that last-known-good value, falling back to `25` only on the first-read edge case. Scope: one field on `PgConfiguration` + the getter.

## Testing

- `cargo test -p documentdb_gateway_core` → **434 passed, 0 failed**, including two new tests (last-known-good on bad read; falls back to 25 with no prior good value) and the existing `validate_max_conn_change` (a real change still flips the key → pool rebuild preserved).
- `rustfmt` and `clippy --workspace --all-targets --all-features -- -D warnings` clean on the repo's pinned toolchain (Rust 1.95).

## Type of Change
- [x] Bug fix

## AI Disclosure
- [x] AI tools were used. **Tool:** Claude (Anthropic). Root-cause tracing, the fix, and the tests were prepared with substantial AI assistance, then reviewed and verified against the source by the author.

## Checklist
- [x] I can explain every change if asked
- [x] Tested locally (`cargo test` — 434 passed)
- [x] Tests added/updated
- [x] Commits signed off (DCO)
- [x] Read CONTRIBUTING.md

---

Signed-off-by: Morchid Chellali <mc@wetransform.to>


AndrewKhoma (changes requested):
thanks for opening this PR, but I need a better explanation on how you've been able to reproduce and execute this state with the running postgresql

pg_documentdb_gw/documentdb_gateway_core/src/configuration/pg_configuration.rs
AndrewKhoma, line 106: this line and the line 102 (`Vec::new()` in `unwrap_or_else`) can be a potential bug you've faced, since we will drop all of the settings, not just the max_connections one in case of any failure instead of returning error

seems wrong to me, since in case of the error it's better to have stale GUCs then none

morch23mj:
Thanks @AndrewKhoma — your inline comment is actually the repro.

**How the state is reached:** it's exactly the `Vec::new()` fallback you flagged. On our deployment a query in the refresh path fails (we see `documentdb_api.get_parameter(...) is not defined` in the logs), so `load_configurations` swaps in a settings map with *all* GUCs dropped — `max_connections` then resolves to the `25` default, which changes the `PgPoolSettings` key and orphans the user pool. Because that query fails on every refresh it's deterministic (~every 300s). To be transparent: we inferred the `max_connections` flip from the symptom + this code path — we didn't capture the GUC value at the moment of failure.

**On the fix:** agreed — dropping every setting on a load error is probably a bug; stale GUCs beat none. I can rework this to keep the previous config on a failed refresh load and drop the now-unnecessary per-getter last-known-good. So — what to do then: **A/** the load returns an error so the refresh skips the swap and keeps the last-good full config (my lean), or **B/** merge onto the current map, upserting only successfully-read rows?

Separately filed #724: the ~300s cadence is itself unexpected — we set `DynamicConfigurationRefreshIntervalSecs` higher but it isn't honored, so the refresh keeps firing at the 300s default.

